PR #76, opened 2015-01-29
Taking out external call on check.

ldecicco-USGS:


lawinslow:
FYI, did a quick survey. @ldecicco-USGS You were correct and I was wrong, as far as I can tell, very few R packages that hit outside services include service calls in their examples that aren't set as "dontrun".

PR #144, opened 2015-11-02
Better encoding and starting switch to dplyr

ldecicco-USGS:


R/constructNWISURL.r
```diff
@@ -244,8 +244,8 @@ constructWQPURL <- function(siteNumber,parameterCd,startDate,endDate){
     suppressWarnings(pCodeLogic <- all(!is.na(as.numeric(parameterCd))))
   } else {
     pCodeLogic <- FALSE
-    parameterCd <- gsub(",","%2C",parameterCd)
-    parameterCd <- URLencode(parameterCd)
+    # parameterCd <- gsub(",","%2C",parameterCd)
```
jread-usgs: why not delete the line?

ldecicco-USGS: because I was careless

ldecicco-USGS: Actually, probably a bunch of commented out stuff I need to delete now that you mention it.

PR #149, opened 2015-11-10
straw dog for access pattern

jread-usgs:
this is just one function, but thoughts @ldecicco-USGS ?


R/setAccess.R
```diff
@@ -0,0 +1,38 @@
...
+  access = match.arg(access, c('public','internal'))
+  
+  if(access=="internal"){
+    access.param = '?Access=3'
+    message('setting access to internal')
+  }else {
+    access.param = '?Access=1'
+  }
+  
+  pkg.env$waterservices = paste0("http://waterservices.usgs.gov/nwis/site/", access.param)
```
ldecicco-USGS: Take off the /site and it's way more useful.

ldecicco-USGS:
Looks good. I was going to see if there were any unintended consequences of setting all the urls to access=3 as a default. It may be that is OK.

ldecicco-USGS:
Even if everyone could use the Access=3 code, this would be a good additional feature though so internal people could see what the public is seeing.

jread-usgs:
would also be nice to check `setAccess('internal')` is being used on the network and warn if not.

R/setAccess.R
```diff
@@ -0,0 +1,40 @@
...
+#' @export
+setAccess = function(access="public"){
+  
+  access = match.arg(access, c('public','internal'))
+  
+  if(access=="internal"){
+    access.param = '?Access=3'
+    message('setting access to internal')
+  }else {
+    access.param = '?Access=1'
```
jread-usgs: should be Access=0

jread-usgs:
@ldecicco-USGS what should I do w/ this PR? Kill it or work this into DR?

ldecicco-USGS:
Go for it. 

My only 'waffle point' was that `drURL` is the first non-exported function, so people can't as easily follow the code. It's also a bit 'magical' because it changes future function behavior (by altering the package environment). But, overall I agree that's less confusing than putting an 'access' argument in all the functions.

jread-usgs:
@ldecicco-USGS I have some new things in here, you have a chance for some feedback? Love it? Hate it?

Main thing is sharing the query builder as a function

man/whatNWISdata.Rd
```diff
@@ -0,0 +1,85 @@
+% Generated by roxygen2: do not edit by hand
```
ldecicco-USGS: So at some point we got an whatNWISdata and whatNWISData (capital D) in here. I'll fix it, but that's why this seems to be a totally new file...because it is.

ldecicco-USGS:
I'm interested in why only 1 of the CI tests failed, it's an actual error in the build on the Travis build.

PR #169, opened 2016-01-26
Switch from RCurl to httr.

ldecicco-USGS:


ldecicco-USGS:
@jiwalker-usgs @jread-usgs @lawinslow 

Could someone have a look at this?

jiwalker-usgs:
closed?

R/importRDB1.r
```diff
@@ -93,10 +99,10 @@ importRDB1 <- function(obs_url, asDateTime=TRUE, convertType = TRUE, tz=""){
                           "America/Phoenix","America/Metlakatla","UTC"))
   }
   
-
   if(file.exists(obs_url)){
     doc <- obs_url
   } else {
+    # doc <- getWebServiceData(obs_url, add_headers(Accept-Encoding = "gzip"))
```
jiwalker-usgs: Cool that this didn't change, probably can remove the commented out code.

ldecicco-USGS: I think the commented-out stuff is not in final release (but I do have to work on being better about not committing that!)

https://github.com/USGS-R/dataRetrieval/blob/master/R/importRDB1.r

ldecicco-USGS:
Yeah, apparently `httr` is going to depend on `xml2` in the very near-term. So, until I can completely switch from `XML` to `xml2`, it's better to stick with `RCurl`.

jread-usgs:
agreed.

PR #174, opened 2016-02-02
Back to a connection.

ldecicco-USGS:


R/importWQP.R
```diff
@@ -57,8 +57,9 @@ importWQP <- function(obs_url, zip=TRUE, tz=""){
   if(!file.exists(obs_url)){
     
     if(zip){
+      message("zip encoding access still in development")
       temp <- tempfile()
-      # options(timeout = 120)
+      options(timeout = 120)
```
jread-usgs: would be nice to allow the user to specify this for known bigger requests?

ldecicco-USGS: If I remember correctly, 120 is as big as you can go. I'll check

ldecicco-USGS: 120 was specifically chosen back in the day for a reason...it's not a random number.

ldecicco-USGS: Well, now I can't find the reason for 120. For the record, I don't think it's quitting if the download hasn't finished in 120 seconds (plenty of mine take longer than that). I think it's if the service hasn't somehow pinged at least 1 bit back in 120 seconds.

jread-usgs: gotcha

ldecicco-USGS: You have led me down a rabbit hole where I'm trying to figure out the difference between R option's timeout and curlSetOpt(timeout = 200)

```
curlSetOpt(timeout = 120)
#vs:
options(timeout = 120)
```
@jiwalker-usgs any insight?

jiwalker-usgs: I have no idea what the difference is here, but I think timeout refers to the readTimeout (how long after no data has been received until stopping).  Both probably do the same thing, but options may apply to other packages.

PR #179, opened 2016-02-22
Second shot at httr

ldecicco-USGS:


ldecicco-USGS:
Well, at least all the tests are consistent. My computer, and the CI tests all fail at the same place. When I run the tests individually, they pass....

PR #199, opened 2016-03-23
Fixed bug in drURL function

veronikayurchuk:
The bug gave an error **"Error in as.vector(x, "character") : 
  cannot coerce type 'externalptr' to vector of type 'character'"** when calling a **whatNWISsites** function

### DIFF
--- a/R/setAccess.R
+++ b/R/setAccess.R
@@ -41,15 +41,14 @@ setAccess = function(access="public"){
     message('setting access to public')
   }
   
-  pkg.env$waterservices = "http://waterservices.usgs.gov/nwis/site/"
-  pkg.env$iv = "http://nwis.waterservices.usgs.gov/nwis/iv/"
-  pkg.env$dv =  "http://waterservices.usgs.gov/nwis/dv/"
-  pkg.env$gwlevels = "http://waterservices.usgs.gov/nwis/gwlevels/"
   options(Access.dataRetrieval = access)
 }
 
 drURL <- function(base.name, ..., arg.list=NULL){
-  
+  pkg.env$waterservices = "http://waterservices.usgs.gov/nwis/site/"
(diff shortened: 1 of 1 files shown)


jread-usgs:
I ran into that bug too, but I don't think these changes address it.

veronikayurchuk:
Hi @jread-usgs !

Function drURL is inside whatNWISsites function. When we call drURL function using for example stateCd="OH",parameterCd=p_code, the output is _"?format=mapper&stateCd=OH&parameterCd=00065"_ instead of _"http://waterservices.usgs.gov/nwis/site/?format=mapper&stateCd=OH&parameterCd=00065"!_

The reason of the problem is in the following code from the drURL function: _return(paste0(pkg.env[[base.name]], '?', queryString))_, it means that_ pkg.env[[base.name]]_, where base.name = 'waterservices' gave "".

One of the possible solution is to define pkg.env$waterservices inside the function drURL. As the result the output of drURL function gave "http://waterservices.usgs.gov/nwis/site/?format=mapper&stateCd=OH&parameterCd=00065"

ldecicco-USGS:
@veronikayurchuk : Could you install from gitHub right now (before your pull request) and verify the bug is still there? I put in a fix to this bug a few days ago to gitHub, and if you are still seeing the error, we'll need to see a reproducible example. So:

```
devtools::install_github("USGS-R/dataRetrieval")
library(dataRetrieval)
OHsites <- whatNWISsites(stateCd="OH",parameterCd="00065")

```

This example works with what is currently up on gitHub. I plan to push it off to CRAN in the next couple of days.

veronikayurchuk:
Hi @ldecicco-USGS !
It works fine!

ldecicco-USGS:
Great. Thanks for checking and digging into the code. I really appreciate it.

PR #222, opened 2016-05-20
WQP query head request 

lindsaycarr:
Find out the number of sites or results that would come back before downloading all of the data.
```r
readWQPdata(..., querySummary=TRUE)
readWQPqw(..., querySummary=TRUE)
```


ldecicco-USGS:
There might be some new notes with R 3.3...I'm going to try again to update, and I'll try to figure out the problem in the importWaterML1 test. But...can you push your changes to the .Rd files? That should get rid of a few notes and I think make the tests you wrote pass.

ldecicco-USGS:
Actually, I just got 3.3 working, and I don't think there are any new notes. So...go ahead and comment out that test, I think it's a NWIS thing (for now). Then, when you update the .Rd files, I *think* this should pass.

lindsaycarr:
this passes checks on my computer...

PR #228, opened 2016-05-27
format head request results

lindsaycarr:


R/readWQPdata.R
```diff
@@ -161,6 +161,13 @@ readWQPdata <- function(..., zip=FALSE, querySummary=FALSE){
   if(querySummary){
     queryHEAD <- HEAD(urlCall)
     retquery <- headers(queryHEAD)
+    countNames <- c('total-site-count', 'nwis-site-count', 'total-result-count', 'nwis-result-count')
```
ldecicco-USGS: So, can you make this a function (a la getWebServices), and then both functions can call it. Then, we only need to edit the one file if we want to modify the returned header output

lindsaycarr: :+1: yep! Should I move the `@importFrom http:HEAD` out of readWQPdata and readWQPqw and just have it for this one new function?

ldecicco-USGS: Yup.

PR #237, opened 2016-06-14
zeroPad NA handling

wdwatkins:


R/zeroPad.r
```diff
@@ -12,15 +12,23 @@
...
 #' correctPCodes <- zeroPad(pCodes,5)
+#' pCodeNA <- c(1,2,NA)
+#' padPCodeNA <- zeroPad(pCodeNA,4)
 zeroPad <- function(x,padTo){
   if(padTo <= 1) return(x)
+  #if(is.na(x) == TRUE) return(x)
   numDigits <- nchar(x)
   padding <- padTo-numDigits
-  padingZeros <- vapply(
+  padNA <- which(is.na(padding) == TRUE)
```
ldecicco-USGS: We never need to use:
```r
blah == TRUE
```
If `blah` is a logical. Just say:
```
padNA <- which(is.na(padding))
```

Or to be more concise:
```r
if(any(is.na(padding))){
  message("There were NAs")
}
```
But...all that being said, I don't think this is `message`-worthy. I think it's a function that the user can expect to put in an `NA` and get out an `NA` without an Error.

R/zeroPad.r
```diff
@@ -12,15 +12,23 @@
...
+    padding[padNA] <- 0
+    message("There were NAs send to zeroPad")
+  }
+  paddingZeros <- vapply(
       X = padding[padding > 0], 
       FUN = function(y) paste(rep("0",y),collapse="",sep=""),
       FUN.VALUE = ""
   )
-  x[padding > 0] <- paste(padingZeros,x[padding > 0],sep="")
+  x[padding > 0] <- paste(paddingZeros,x[padding > 0],sep="")
```
ldecicco-USGS: I know this is my old code, but anytime you have the function:

```r
paste(x, sep="")
```
You can instead use:
```r
paste0(x)
```
saves a lot of `sep=""` scattered all over the code.

PR #239, opened 2016-06-14
Date with no day fix, example, test

wdwatkins:


R/readNWISunit.r
```diff
@@ -387,13 +387,19 @@ readNWISmeas <- function (siteNumbers,startDate="",endDate="", tz="", expanded=F
 #' sites <- c("434400121275801", "375907091432201")
 #' data2 <- readNWISgwl(sites, '','')
 #' data3 <- readNWISgwl("420125073193001", '','')
+#' data4 <- readNWISgwl("425957088141001", startDate = "1980-01-01") #handling of data where date has no day
```
ldecicco-USGS: Need to get the updated .Rd file in this pull request.

R/readNWISunit.r
```diff
@@ -374,7 +374,7 @@ readNWISmeas <- function (siteNumbers,startDate="",endDate="", tz="", expanded=F
 #' url \tab character \tab The url used to generate the data \cr
 #' queryTime \tab POSIXct \tab The time the data was returned \cr
 #' comment \tab character \tab Header comments from the RDB file \cr
-#' siteInfo \tab data.frame \tab A data frame containing information on the requested sites \cr
+#' siteInfo \tab xdata.frame \tab A data frame containing information on the requested sites \cr
```
ldecicco-USGS: typo?

R/readNWISunit.r
```diff
@@ -387,13 +387,19 @@ readNWISmeas <- function (siteNumbers,startDate="",endDate="", tz="", expanded=F
...
 #' data2 <- readNWISgwl(sites, '','')
 #' data3 <- readNWISgwl("420125073193001", '','')
+#' data4 <- readNWISgwl("425957088141001", startDate = "1980-01-01") #handling of data where date has no day
 #' }
 readNWISgwl <- function (siteNumbers,startDate="",endDate="", convertType = TRUE){  
   
   url <- constructNWISURL(siteNumbers,NA,startDate,endDate,"gwlevels",format="tsv")
   data <- importRDB1(url,asDateTime=TRUE, convertType = convertType)
 
   if(nrow(data) > 0){
```
ldecicco-USGS: Let's add `convertType` when we are doing the conversions, so the if statement would be:

```r
if(nrow(data) >0){
  if(convertType){
     #...all the date conversion lines...
  }
  siteInfo <- #...still want to do this stuff whether or not the user chooses `convertType`
}

```

PR #242, opened 2016-06-21
Added stats service to readNWISdata

wdwatkins:


ldecicco-USGS:
Looks great!

jread-usgs:
Increment minor version number?

ldecicco-USGS:
yeah, I'll do that with a push to gran too

PR #244, opened 2016-06-22
Added NA handling in siteNumbers for readNWISstat

wdwatkins:


R/readNWISunit.r
```diff
@@ -472,4 +480,77 @@ readNWISstat <- function(siteNumbers, parameterCd, startDate = "", endDate = "",
...
+  url <- constructUseURL(years,stateAB,county)
+  data <- importRDB1(url,convertType=convertType)  
+  
+  #distinguish if total country data arriving in named rows
+  if(is.null(county) && is.null(stateAB)){
+    cmmnt <- comment(data)
+    data <- t(data)
+    colnames(data) <- data[1,]
+    data <- as.data.frame(data[-1,],stringsAsFactors=FALSE)
+    data <- cbind(Year=as.integer(str_sub(rownames(data),2)),data)
```
ldecicco-USGS: yeah, this will cause errors without adding a `stringr` as a new dependency. In general, I've been trying to limit the dependencies as much as possible. Let's see if we can do it in base

ldecicco-USGS: Add `reshape = TRUE` (by default) argument to let the user decide if they want it transformed (or maybe call it transform)

R/readNWISunit.r
```diff
@@ -472,4 +480,77 @@ readNWISstat <- function(siteNumbers, parameterCd, startDate = "", endDate = "",
...
+#' 
+#' #All national-scale data
+#' national <- readNWISuse()
+#' 
+#' #Washington, DC data
+#' dc <- readNWISuse(stateAB = "DC")
+#' 
+#' }
+ 
+readNWISuse <- function(years="ALL",stateAB=NULL,county=NULL,convertType=TRUE){
```
ldecicco-USGS: Change to:
```
function(stateAB, county, years="ALL", convertType=TRUE){
```
Then remove the stop bit

R/constructNWISURL.r
```diff
@@ -305,3 +305,27 @@ constructWQPURL <- function(siteNumber,parameterCd,startDate,endDate,zip=FALSE){
...
+      if(!is.null(county)){
+        if(length(county) > 1) {county <- paste(county,collapse="%2C")}
+          baseURL <- paste0(baseURL,"&wu_area=county&wu_county=",county)
+      } else{
+          baseURL <- paste0(baseURL,"&wu_area=State%20Total")
+        }
+      }
+    years <- paste(years, collapse="%2C")
+    retURL <- paste0(baseURL,"&wu_year=",years,"&wu_category=ALL")
+
```
ldecicco-USGS: Could use `URLencode`

R/constructNWISURL.r
```diff
@@ -305,3 +305,27 @@ constructWQPURL <- function(siteNumber,parameterCd,startDate,endDate,zip=FALSE){
   return(url)
 
 }
+
+#' Construct URL for NWIS water use data service
+#' 
+#' 
+
+constructUseURL <- function(years,stateAB,county){
```
ldecicco-USGS: Let's use `stateCd`:
stateCdLookup

R/readNWISunit.r
ldecicco-USGS: Actually, change stateAB to stateCd,

R/constructNWISURL.r
wdwatkins: I'm unsure about using stateCd as a variable name, when I need to call my stateCdLookup function that uses the stateCd data frame.  I've gotten errors with it trying to look up things in the variable, but I can't seem to reproduce it consistently.

R/readNWISunit.r
wdwatkins: I'm unsure about using stateCd as a variable name here, since I need to call my stateCdLookup function that uses the stateCd data frame. I've gotten errors with it trying to look up things in the variable, but I can't seem to reproduce it consistently.

ldecicco-USGS: Hmmm....OK, maybe just change it to `state`? I like using the `stateCdLookup` function, and I like that the argument name doesn't imply it has to be just the abbreviation

R/constructNWISURL.r
```diff
@@ -305,3 +313,37 @@ constructWQPURL <- function(siteNumber,parameterCd,startDate,endDate,zip=FALSE){
...
+
+#' Construct URL for NWIS water use data service
+#' 
+#' Reconstructs URLs to retrieve data from here: \url{http://waterdata.usgs.gov/nwis/wu}
+#' 
+#' @param years integer Years for data retrieval. Must be years ending in 0 or 5, or "ALL", which retrieves all available years.
+#' @param stateCd could be character (full name, abbreviation, id), or numeric (id)
+#' @param countyCd could be numeric (County IDs from countyCdLookup) or character ("ALL") 
+#' @return url string
+#' @export
```
ldecicco-USGS: The export here doesn't seem to make it to the NAMESPACE file.

ldecicco-USGS:
(Expanding on the line note)
One error in travis is:
```
Error: could not find function "constructUseURL"
```
I see you have @export in the roxygen part, but it also needs to update the NAMESPACE. You might need to go to the Build tab, then "Configure Build Tools", then "Generate Roxygene" or something like that.

PR #253, opened 2016-07-13
fixed additional importWaterML2 bugs, added tests

wdwatkins:


R/importWaterML2.r
```diff
@@ -53,7 +53,7 @@ importWaterML2 <- function(obs_url, asDateTime=FALSE, tz=""){
     rawData <- obs_url
     doc <- xmlTreeParse(rawData, getDTD = FALSE, useInternalNodes = TRUE)
   } else {
-    doc <- xmlTreeParse(getWebServiceData(obs_url), getDTD = FALSE, useInternalNodes = TRUE)
+    doc <- xmlTreeParse(obs_url, getDTD = FALSE, useInternalNodes = TRUE)
```
ldecicco-USGS: We'll want to think about how to incorporate the features of `getWebServiceData` (user agents, header responses, etc). In the meantime, if we aren't using it, this if/else statement seems useless.

wdwatkins: Oh yeah, didn't catch that.  getWebServiceData was causing errors for some reason on some URLS.

ldecicco-USGS: Yeah, almost certainly because the content-type doesn't match the list we have

ldecicco-USGS: You could try adding a regx to 
https://github.com/USGS-R/dataRetrieval/blob/master/R/getWebServiceData.R#L40

any content-type with "xml" gets sent to `xmlcontent `

PR #267, opened 2016-08-09
importWaterML1 via xml2

wdwatkins:


wdwatkins:
failing the new two data descriptors test

R/importWaterML1.r
```diff
@@ -97,356 +95,194 @@
...
-#' imporFile <- importWaterML1(fullPath,TRUE)
+#' importFile <- importWaterML1(fullPath,TRUE)
 #'
-importWaterML1 <- function(obs_url,asDateTime=FALSE, tz=""){
-  
-  if(file.exists(obs_url)){
-    rawData <- obs_url
-    returnedDoc <- xmlTreeParse(rawData, getDTD = FALSE, useInternalNodes = TRUE)
+
+importWaterML1 <- function(input,asDateTime=FALSE, tz=""){
```
ldecicco-USGS: Even though I think it's good to let the `obs_url` take in a raw XML....I think it might be a bad idea to change the name of the argument. If people have code specified as:
```
importWaterML1(obs_url=fullPath,TRUE)
```
it's going to break scripts. So, can we keep the new functionality but change the name back?

tests/testthat/tests_userFriendly_fxns.R
```diff
@@ -31,8 +31,8 @@ test_that("Unit value data returns correct types", {
   endDate <- "2012-07-17"
   dd_2 <- readNWISuv(site, pCode, startDate, endDate)
   expect_true(all(names(dd_2) %in% c("agency_cd","site_no",                   
-                                 "dateTime","X_.YSI.6136.UP._63680_00011",   
-                                 "X_YSI.6136.DOWN_63680_00011","X_.YSI.6136.UP._63680_00011_cd",
+                                 "dateTime","X_YSI.6136.UP_63680_00011",
```
ldecicco-USGS: Hmm...while I like this better...it will break people's scripts again. Let me think on it.

R/importWaterML1.r
wdwatkins: Is this also a concern for `importWaterML2`?

ldecicco-USGS: I don't think it's a concern that we'll break someone's script (I don't think anyone's really using it). The only thing might be to keep it consistent with `importWaterML1` so someone could conceivably swap out the functions without much hassle.

Issue #268, opened 2016-08-10
xml2 updates


ldecicco-USGS:
- [x] Incorporate `getWebServiceData` function
- [x] Add option for "raw" data
- [x] Increase test coverage for increased risk

ldecicco-USGS:
Possibly, break out `importWaterML1` to more internal functions to make testing easier

wdwatkins:
Also related to yesterday's discussion - I tested an invalid xml file, and read_xml fails nicely.  Web service data is no longer going through read_xml with `getWebServiceData`, but I assume it would be caught.

ldecicco-USGS:
- [ ] Sort by site_no
- [ ] Change the names in the statisticInfo attribute to remain backwards compatible:
- [ ] Move tz_cd column to far right
- [ ] For the siteInfo attribute, `dec_lat_va` and `dec_lon_va` need to be numeric
- [ ] get the "no value" value
- [ ] replace the "no value" value with NA
- [ ] change the noDataValue to NA in variableInfo attribute

wdwatkins:
@ldecicco-USGS what do you mean by sort by site_no?

ldecicco-USGS:
I think old fashioned `dataRetrieval` comes back sorted by increasing site_no then increasing dates. So, for:

```
bBoxEx <- readNWISdata(bBox=c(-83,36.5,-81,38.5), parameterCd="00010")
head(bBoxEx[,1:3]
  agency_cd  site_no   dateTime
1      USGS 03178000 1998-09-30
2      USGS 03185400 1998-09-30
3      USGS 03191500 1998-09-10
4      USGS 03193000 1998-09-30
5      USGS 03193742 1992-09-30
6      USGS 03193770 1977-06-30
```
Whereas I think this version sorts by decreasing site_no(?) and decreasing dates:
```
bBoxEx <- readNWISdata(bBox=c(-83,36.5,-81,38.5), parameterCd="00010")
head(bBoxEx[,1:3]
  agency_cd  site_no   dateTime
1      USGS 03527000 1950-09-30
2      USGS 03527000 1967-09-30
3      USGS 03526000 1998-09-30
4      USGS 03524500 2016-08-23
5      USGS 03490000 1970-12-31
6      USGS 03490000 2016-08-23
```

wdwatkins:
Oh.  This version probably just comes back in whatever order the sites were given in.

PR #269, opened 2016-08-12
importWaterML2 to xml2

wdwatkins:
Just importing the basic info right now - time, value, qualifier, and identifier.  Will be fleshed out for NGWMN soon


R/importWaterML1.r
```diff
@@ -97,356 +95,194 @@
...
-            
-            numChar <- nchar(datetime)
-            
-            datetime <- parse_date_time(datetime, c("%Y","%Y-%m-%d","%Y-%m-%dT%H:%M",
-                                                    "%Y-%m-%dT%H:%M:%S","%Y-%m-%dT%H:%M:%OS",
-                                                    "%Y-%m-%dT%H:%M:%OS%z"), exact = TRUE)
-            
-            if(any(numChar < 20) & any(numChar > 16)){
-              
-              offsetLibrary <- data.frame(offset=c(5, 4, 6, 5, 7, 6, 8, 7, 9, 8, 10, 10, 0),
```
ldecicco-USGS: @wdwatkins ...are you getting around this conversion somewhere else?

ldecicco-USGS:
One thing that's different is that the "production" version of dataRetrieval (that is on CRAN) changes the "no value" values to NA. So that the user doesn't need to check if the value == -999999. 

So, you'll want to get the "no value" value here:
https://github.com/USGS-R/dataRetrieval/blob/master/R/importWaterML1.r#L161
Then change it to NA something like:
https://github.com/USGS-R/dataRetrieval/blob/master/R/importWaterML1.r#L422

ldecicco-USGS:
For the siteInfo attribute, `dec_lat_va` and `dec_lon_va` need to be numeric.

ldecicco-USGS:
For the variableInfo attribute, if you convert the "no value" value to NA, then change the noDataValue to NA as well.

ldecicco-USGS:
Column order is different. Most specifically the tz_cd column should go to the end. In the older versions, the values are lumped together, and the cd's are lumped together. I prefer them to be lumped by parameter code (as your new version has)...but we have to be prepared for people's scripts breaking. I'd say leave your order as is, but move the tz_cd to the far right.

For example:

```
bBoxEx <- readNWISdata(bBox=c(-83,36.5,-81,38.5), parameterCd="00010")
```

The CRAN version is:
```
[1] "agency_cd"        "site_no"          "dateTime"         "X_00010_00001"   
 [5] "X_00010_00002"    "X_00010_00003"    "X_00010_00011"    "X_00010_00008"   
 [9] "X_00010_00001_cd" "X_00010_00002_cd" "X_00010_00003_cd" "X_00010_00011_cd"
[13] "X_00010_00008_cd" "tz_cd"   
```
This version is:
```
[1] "agency_cd"        "site_no"          "dateTime"         "X_00010_00001"   
 [5] "X_00010_00001_cd" "tz_cd"            "X_00010_00008"    "X_00010_00008_cd"
 [9] "X_00010_00003"    "X_00010_00003_cd" "X_00010_00002"    "X_00010_00002_cd"
[13] "X_00010_00011"    "X_00010_00011_cd"
```

ldecicco-USGS:
Let's change the names in the statisticInfo attribute to remain backwards compatible:
```
[1] "statisticName" "statisticCd"  
```
instead of:
```
[1] "statCd"  "stat_nm"
```

ldecicco-USGS:
Moving to #268

wdwatkins:
restarting travis

wdwatkins:
Try again

PR #272, opened 2016-08-24
NGWMN getObservation

wdwatkins:
Multi-site observation requests work, and attach site metadata also.  The featureOfInterest retrieval isn't ready to be exposed though I think, as it doesn't support bbox yet.  

Also time zone offsets are appearing for data that has real times.  We'll still need to decide how to deal with times that are just `12:00:00` or `00:00:00`, since they do not.  We could make an option to drop times and just return those as dates?


R/importNGWMN_wml2.R
```diff
@@ -0,0 +1,125 @@
...
+  }
+  return(mergedDF)
+  
+}
+
+#replace blank cells with NAs
+removeBlanks <- function(input){
+  input <- sapply(input, function(f){is.na(f)<-which(f == '');f})
+  return(input)
+}
```
ldecicco-USGS: Is this any different than:
```
input[input == ""] <- NA
```

wdwatkins: 😬

wdwatkins:
travis restart

wdwatkins:
again...

wdwatkins:
I've added the two user-friendly functions, and hidden `readNGWMNdata` for now since the other two have all the same functionality at the moment, and a few more things need to be built out for `NGWMNdata` to really do anything different

Issue #274, opened 2016-08-26
update whatNWISsites to xml2


ldecicco-USGS:
and add some coverage :grimacing:

ldecicco-USGS:
Here's the url:
http://waterservices.usgs.gov/nwis/site/?format=mapper&stateCd=OH&parameterCd=00665

For:
```
siteListPhos <- whatNWISsites(stateCd="OH",parameterCd="00665")
```

PR #277, opened 2016-08-31
whatNWISsites to xml2

wdwatkins:
Now includes logical column to denote colocated sites.  The old version wasn't returning them at all.  Also added tests


wdwatkins:
#274

Issue #279, opened 2016-09-01
Column name weirdness


ldecicco-USGS:
The latest version breaks:

https://github.com/USGS-OWI/owi-blog/blob/master/content/statsServiceMap.Rmd

I verified that the CRAN version works.

ldecicco-USGS:
Also some funkyness in ts-colin-discharge:

https://github.com/USGS-OWI/owi-blog/pull/115/files

wdwatkins:
The discharge plot is off because the rows aren't being ordered by data.  Their must be some out-of-order points for that site

wdwatkins:
#280 

The stats map post was breaking because there were some goofy sites in ND with different method descriptors, and attribute name mismatches

ldecicco-USGS:
There's still a problem with ND query:

dataRetrieval from CRAN:
```r
install.packages("dataRetrieval", repos = "https://cloud.r-project.org")
storm.date <- "2016-07-12"
st <- "ND"
stDV <- renameNWISColumns(readNWISdata(service="dv",
                              parameterCd="00060",
                              stateCd = st,
                              startDate = storm.date,
                              endDate = storm.date))

> nrow(stDV)
[1] 100

#vs
detach("package:dataRetrieval", unload=TRUE)
devtools::install_github("USGS-R/dataRetrieval")
library(dataRetrieval)
stDV <- renameNWISColumns(readNWISdata(service="dv",
                              parameterCd="00060",
                              stateCd = st,
                              startDate = storm.date,
                              endDate = storm.date))

Error: 'dateTime' column not found in rhs, cannot join

```

PR #282, opened 2016-09-06
update version tag

wdwatkins:


wdwatkins:
another bug fix for #279 coming

R/importWaterML1.r
```diff
@@ -244,7 +244,12 @@ importWaterML1 <- function(obs_url,asDateTime=FALSE, tz=""){
           obsDF <- full_join(obsDF, valParentDF, by = c("dateTime","tz_cd"))
         }
       }else{
-        obsDF <- data.frame()
+        #need column names for joining later
+        obsDF <- data.frame(dateTime=character(0), tz_cd=character(0), stringsAsFactors = FALSE)
+        if(asDateTime){
+          obsDF$dateTime <- as.POSIXct(obsDF$dateTime)
```
ldecicco-USGS: This probably needs (or does it have somewhere else?)

```
  if(asDateTime){
    if(tz != ""){
      attr(obsDF$dateTime, "tzone") <- tz
      obsDF$tz_cd <-tz
    } else {
      attr(obsDF$dateTime, "tzone") <- "UTC"
      obsDF$tz_cd <-"UTC"
    }
}
```

wdwatkins: It would be set earlier when there is actual data.  This is just creating the columns so there isn't an error when you merge an empty data frame.  Would time zones still matter on an empty data frame?

wdwatkins: I think it would maintain the time zone from the data, based on my little experiments.  But wouldn't hurt to add it here also.  

time zone is set in the parse_date_time section (line 212)

PR #288, opened 2016-11-17
WQP to https

wdwatkins:
The build is going to fail, NWIS stat code thing is making tests fail and seems NGWMN is down


jread-usgs:
Could/should this be a different branch than the NGWMN work, since we don't know when that can be completed, and this may need to go to CRAN first?

ldecicco-USGS:
Yeah, that would be ideal.

wdwatkins:
I have a separate branch on my fork with all the current NGWMN stuff on it (beyond what is on master here).  So we could simply delete the NGWMN stuff here.  Would we want to then keep a copy of the NGWMN branch on the canonical?

jread-usgs:
yes, creating a new branch on canonical would be the way to go

PR #292, opened 2016-12-06
http -> https

ldecicco-USGS:
@jiwalker-usgs or @jread-usgs  or @wdwatkins 

The tests take awhie because each one has to fail first, then got get the real data. 

I didn't change any of the URLs in the documentation because I think CRAN actually checks those, and they'd all be broken. So, we'll have to do (I think), another release to get those all working.


wdwatkins:
looks ok to me

PR #300, opened 2017-01-20
https and tz

ldecicco-USGS:


ldecicco-USGS:
@aappling-usgs this should be a pretty straightforward review. I added you mostly to get a 2nd pair of eyes on the timezone. I changed it so ALL legal timezones are available...but the documentation will still focus on the main US ones. Let me know if you like/dislike the idea

R/importRDB1.r
```diff
@@ -86,11 +86,9 @@
...
   
   if(tz != ""){
-    tz <- match.arg(tz, c("America/New_York","America/Chicago",
-                          "America/Denver","America/Los_Angeles",
-                          "America/Anchorage","America/Honolulu",
-                          "America/Jamaica","America/Managua",
-                          "America/Phoenix","America/Metlakatla","UTC"))
+    tz <- match.arg(tz, OlsonNames())
+  } else {
+    tz <- "UTC"
```
aappling-usgs: I agree this is a good value to use as a default. How about including it in the function declaration rather than reassigning its value here? So back on line 86: 
```r
 importRDB1 <- function(obs_url, asDateTime=TRUE, convertType = TRUE, tz="UTC"){
```

R/importWQP.R
```diff
@@ -47,11 +47,9 @@
...
   
   if(tz != ""){
-    tz <- match.arg(tz, c("America/New_York","America/Chicago",
-                          "America/Denver","America/Los_Angeles",
-                          "America/Anchorage","America/Honolulu",
-                          "America/Jamaica","America/Managua",
-                          "America/Phoenix","America/Metlakatla"))
+    tz <- match.arg(tz, OlsonNames())
+  } else {
+    tz <- "UTC"
```
aappling-usgs: same as above

R/importWaterML1.r
```diff
@@ -189,9 +187,8 @@ importWaterML1 <- function(obs_url,asDateTime=FALSE, tz=""){
 
       nObs <- length(values)
       qual <- xml_attr(obs,"qualifiers")
-      if(all(is.na(qual))){
-        noQual <- TRUE
-      }else{noQual <- FALSE}
+      
+      noQual <- all(is.na(qual))
```
aappling-usgs: :+1:

R/importRDB1.r
```diff
@@ -9,7 +9,7 @@
 #' @param asDateTime logical, if \code{TRUE} returns date and time as POSIXct, if \code{FALSE}, Date
 #' @param tz character to set timezone attribute of datetime. Default is an empty quote, which converts the 
 #' datetimes to UTC (properly accounting for daylight savings times based on the data's provided tz_cd column).
-#' Possible values to provide are "America/New_York","America/Chicago", "America/Denver","America/Los_Angeles",
+#' Possible values to provide include "America/New_York","America/Chicago", "America/Denver","America/Los_Angeles",
```
aappling-usgs: Maybe also point people to OlsonNames and/or https://en.wikipedia.org/wiki/List_of_tz_database_time_zones for the complete list? I always have to search around for those. 

To maintain emphasis on the America options, you could even change the above to say that you *recommend* those.

And maybe add UTC to affirm it as one of the recommended options?

```
Recommended values to provide are "UTC","America/New_York","America/Chicago", 
"America/Denver","America/Los_Angeles","America/Anchorage","America/Honolulu",
"America/Jamaica","America/Managua","America/Phoenix", or "America/Metlakatla".
See https://en.wikipedia.org/wiki/List_of_tz_database_time_zones for a complete list of possibilities.
```

aappling-usgs:
:+1:

Issue #284, opened 2016-10-21
support list methods for function calls


jread-usgs:
example:

``` r
readWQPdata(siteid="USGS-04024315",characteristicName="pH")
```

is a thing, but often we want to build up the list and use it:

``` r
readWQPdata(list(siteid="USGS-04024315",characteristicName="pH"))
```

This ☝️ wouldn't be supported because the signature for `readWQPdata` is `(..., zip = FALSE, querySummary = FALSE)`, but could be done by adding a new arg so it doesn't break people's scripts:

``` r
readWQPdata(query = list(siteid="USGS-04024315",characteristicName="pH"))
```

and be smart enough to deal with conflicts and additional params: 

``` r
readWQPdata(query = list(siteid="USGS-04024315",characteristicName="pH"), countycode="US:55:025", characteristicName="temperature")
```

jread-usgs:
Actually, wondering if 
```r
readWQPdata(x, ..., zip = FALSE, querySummary = FALSE)
```
would break things. The `...` have to be named, otherwise they don't end up in the query, right? so if you had
`readWQPdata.list`, `readWQPdata.character`, `readWQPdata.missing` etc...they may not break anything for others.

jread-usgs:
My existing workflow that would be improved by this (and easier for others to understand):
```r
secchi.names = c("Depth, Secchi disk depth", "Depth, Secchi disk depth (choice list)", "Secchi Reading Condition (choice list)", "Secchi depth", "Water transparency, Secchi disc")
char.names = sapply(secchi.names, list)
names(char.names) <- rep('characteristicName', length(char.names))
args <- list('startDateLo' = startDateLo, 'startDateHi' = startDateHi, statecode=fip)
args <- append(args, char.names)
wqp.data <- do.call(retryWQP, args)
```

PR #301, opened 2017-01-20
Allow lists in readXdata

ldecicco-USGS:


R/readNWISdata.r
```diff
@@ -95,16 +95,26 @@
 #'                           parameterCd="00065")
 #' allDailyStats <- readNWISdata(site=c("03111548"),
 #'                               service="stat",
-#'                               statReportType="daily")
```
ldecicco-USGS: FYI...I think this is just an old typo.

ldecicco-USGS:
@jread-usgs this is in response to #284

R/readWQPdata.R
```diff
@@ -96,12 +96,25 @@
...
+#'                  "Water transparency, Secchi disc")
+#' args <- list('startDateLo' = startDate, 
+#'              'startDateHi' = "2013-12-31", 
+#'               statecode="WI", 
+#'               characteristicName=secchi.names)
 #' 
-#'                         
+#' wqp.data <- readWQPdata(args)                  
 #' }
 readWQPdata <- function(..., zip=FALSE, querySummary=FALSE){
```
jread-usgs: I link this would be useful if we can hammer out some of the details and figure out what the expected pattern should be. 

Questions:
1) do we accept more than one list in `...`?
2) can you override arguments outside of the list? (e.g., what happens when I specify a value for `zip` in the list?
3) can non-list args be included in `...` and what happens if not? (i.e., if we have one list in `...` and the rest args, does it fail gracefully and let you know why?)

I guess I was thinking this would have a list, not be all of none on `...` being lists. A couple of challenges that we might run into with this pattern:
```r
args <- list('startDateLo' = startDate, 
              'startDateHi' = "2013-12-31")
wqp.data <- readWQPdata(args, statecode="WI", characteristicName=secchi.names)
```
or
```r
args <- list('startDateLo' = startDate, 
              'startDateHi' = "2013-12-31", zip = FALSE)
wqp.data <- readWQPdata(args) 
```
The main use-case I would have is having all of the args in a list and using `do.call`, but I think `zip` and `querySummary` would be blocked from that pattern in this implementation.

ldecicco-USGS: So in a comment I have "TODO:" and it calls out extracting the defaults (`zip` for WQP, `convertType` and something else for NWIS), meaning if this PR was a good start, it would probably comply with your last "main use-case" statement when that TODO was done. 

I definitely had not considered multiple lists, and did assume it was going to be all-or-none (except for the default non-`...` args)....I'd need to wrap my head a bit around how to deal with that.

lindsaycarr:
Possibly use `missing` to see if user-defined or default was used

```r
myfunc <- function(y="default"){
  missing(y)
}
myfunc(y="not y")
[1] FALSE
myfunc()
[1] TRUE
myfunc(y="default")
[1] FALSE
```

ldecicco-USGS:
Then we're saying user-supplied arguments trump list. But list trumps default

R/readWQPdata.R
```diff
@@ -96,12 +97,50 @@
...
+    for(i in c("zip","querySummary")){
+      if(do.call(missing, list(i)) & i %in% names(matchReturn)){
+        assign(i, matchReturn[[i]])
+      }
+      
+      if(i %in% names(matchReturn)){
+        matchReturn <- matchReturn[-which(names(matchReturn) %in% i)]
+      }
+    }
+  }
```
lindsaycarr: use some type of append_replace (like in gsplot) where zip and querySummary are added to the list. Need more logic for building URL in this method, but eliminates need for `assign()`

ldecicco-USGS:
@jiwalker-usgs can you have a look at the code? It was working before the WQP went down. Once that's I'll double-check the examples and tests are still passing.

jiwalker-usgs:
Looks good to me, will the method signature change effect anyone?

ldecicco-USGS:
I don't think so. If someone was leaving `service` blank before, that will still default to "dv". Same with `zip`. The idea was that it should all still work as before, and the tests make me think that's true. But a second set of eyes looking for a way that breaks now and didn't before would be good.

Issue #246, opened 2016-07-01
add water year


ldecicco-USGS:

ldecicco-USGS:
Add this as a helper function.

lindsaycarr:
water year script example:
```r
dates <- seq(as.Date("2010-01-01"),as.Date("2012-12-31"), by="months")
PosLT <- as.POSIXlt(dates)
waterYear <- PosLT$year+1900
waterYear[PosLT$mon >= 9] <- waterYear[PosLT$mon >= 9]+1
```

PR #319, opened 2017-02-16
addWaterYear!

lindsaycarr:
Easily add a waterYear column to your dataRetrieval calls to address #246:

```r
instFlow <- readNWISdata(sites="05114000", service="iv", 
                   parameterCd="00060", 
                   startDate="2014-05-01T00:00Z",endDate="2014-05-01T12:00Z") %>% 
    addWaterYear
```

- [X] NWIS formatted df, `dateTime` column name
- [x] NWIS formatted df, `Date` column name
- [x] WQP formatted df, lots of date/time columns + fix test


ldecicco-USGS:
Ohhhh shoot, that's why you were asking about Rd files...I thought you just meant the one for your function. Now it's a race to see who's PR gets done first, and who needs to deal with 43 merge conflicts. Or...is git smart enough to handle that?

R/addWaterYear.R
```diff
@@ -0,0 +1,41 @@
...
+  waterYear <- calYear
+  waterYear[calMon >= 10] <- calYear[calMon >= 10] + 1 
+  
+  # add water year vector as new column
+  wyData <- rawData
+  wyData$waterYear <- waterYear
+  
+  # move waterYear so that it is always comes right after dateTime
+  dateTime_i <- which(names(wyData) == "dateTime")
+  wyData <- select(wyData, 1:dateTime_i, waterYear, everything())
```
ldecicco-USGS: I'd be shocked if you didn't need to add

```
#' importFrom dplyr everything
```

lindsaycarr: ah thanks - didn't see that come up as a warning...hmmm

lindsaycarr:
```r
# WQP test ("ActivityStartDateTime" col): 
pHData <- readWQPdata(siteid="USGS-04024315",characteristicName="pH")
pHData <- addWaterYear(pHData) 
# currently throws an error:
Error in addWaterYear(pHData) : 
  specified date column does not exist in supplied data frame

# readNWISdv test ("Date" col):
rawDailyQ <- readNWISdv(siteNumber = '04085427', parameterCd = '00060', 
                        startDate = '2012-01-01', endDate = '2012-06-30')
rawDailyQ <- addWaterYear(rawDailyQ)
```

ldecicco-USGS:
The WQP functions don't come back with a column called dateTime, there are 2 ActivityStartTime and ActivityEndDate (and we also add ActivityEndDateTime and ActivityStartDateTime), since water quality measurements are not necessarily instantaneous measurements (like, a filter might be sitting in the water for days).

lindsaycarr:
So which of those column names should I add as the default to use for any WQP data?

R/addWaterYear.R
```diff
@@ -0,0 +1,83 @@
...
+  
+  return(wyData)
+}
+
+#' extract WY from a date
+#' 
+#' @param dateVec vector of dates as character ("YYYY-DD-MM"),
+#' Date, or POSIXct. Numeric does not work.
+#' 
+#' @keywords internal
```
ldecicco-USGS: I think we should expose this. It could be useful to people. Maybe they've altered their data frame or have their own spreadsheet.

ldecicco-USGS: I've given this code out enough that it warrants a function....

R/addWaterYear.R
```diff
@@ -0,0 +1,83 @@
...
+  # only allow WY to be added if there is an appropriate date column
+  if(all(!allowedDateColNames %in% names(rawData))){
+    stop("specified date column does not exist in supplied data frame")
+  }
+  
+  # set the name of the date column(s) to use for calculating the WY
+  dateColName <- names(rawData)[names(rawData) %in% allowedDateColNames]
+  
+  # set up new df that will include water year columns
+  wyData <- rawData
```
ldecicco-USGS: What's the point of this? On the off-chance that the data frame is HUGE, there will be 2 copies of it...I assume that's not great for performance. I'm not an expert on how R allocates memory though... @jiwalker-usgs thoughts?

lindsaycarr: it's literally just so I don't return `rawData` at the end. That bothered me since it was no longer "raw" - I altered it :)

lindsaycarr: I've got no issue whatsoever with deleting this and just using `rawData` the whole time

R/addWaterYear.R
```diff
@@ -0,0 +1,83 @@
...
+  
+  return(wyData)
+}
+
+#' extract WY from a date
+#' 
+#' @param dateVec vector of dates as character ("YYYY-DD-MM"),
+#' Date, or POSIXct. Numeric does not work.
+#' 
+#' @export
```
ldecicco-USGS: Then need add some examples and add a bit better description (for example, I think water year starting at October is what the USGS has decided is a water year. Other parts of the world may choose other months....but this function is specific to our definition.

R/addWaterYear.R
```diff
@@ -0,0 +1,83 @@
...
+#' dataTemp <- readNWISdata(stateCd="OH",parameterCd="00010", service="dv")
+#' dataTemp <- addWaterYear(dataTemp)
+#' 
+#' pHData <- readWQPdata(siteid="USGS-04024315",characteristicName="pH")
+#' pHData <- addWaterYear(pHData)
+#' }
+addWaterYear <- function(rawData){
+  
+  allowedDateColNames <- c("dateTime", "Date", "ActivityStartDate", "ActivityEndDate")
+
```
ldecicco-USGS: What about:
```
allowedDateColNames <- c("waterYear","waterYear","ActivityStartWaterYear","ActivityEndWaterYear")
names(allowedDateColNames ) <- c("dateTime", "Date", "ActivityStartDate", "ActivityEndDate")
```

Then, down on line 42:
```
for(dateCol in names(dateColName)){
   dateColWY <- dateColName[dateCol]


```

R/addWaterYear.R
```diff
@@ -36,35 +38,45 @@ addWaterYear <- function(rawData){
...
-  # set up new df that will include water year columns
-  wyData <- rawData
-  
   for(dateCol in dateColName){
-    dateColWY <- paste0(dateCol, "WY")
+    dateColWY <- allowedWYColNames[dateCol]
     
     # if this WY column already exists, do not add another
-    if(dateColWY %in% names(wyData)){
+    if(dateColWY %in% names(rawData)){
```
ldecicco-USGS: I think you could get this logic in line 39 too (?):

Something like (untested!!!!):
```
dateColName <- names(rawData)[names(rawData) %in% allowedDateColNames]
dateColNames <- dateColNames[!(dateColNames %in% names(rawData))]
```
or put an `&` directly in 39. 

In this case, we can totally keep that `if/next`, since at most, this loop is happening 4 times. In other loops that could be happening for 100's or 1000's or more times, I'm pretty sure a `if/next` statement is pretty slow compared to doing it ahead of time outside of the loop. That's why I always try to see if I can get them out just as best practice.

lindsaycarr: That's a great point, thanks. I'll try to work that in before the loop!

lindsaycarr: Wait, a little confused here. I am already doing the above lines you provided with `dateColName <- names(rawData)[names(rawData) %in% allowedDateColNames]` which only returns the column name to loop over that exists in `rawData`. I need to replace the if/next which deals with the WY col names. So following the same pattern-ish but for WY col names.

Issue #229, opened 2016-05-27
implement POST retrieval for WQP


ldecicco-USGS:

ldecicco-USGS:
Linger question: What to do with URL attribute that's too long

PR #322, opened 2017-02-17
POST if url is too long

jiwalker-usgs:
May conflict with other work going on concurrently.  2048 length is arbitrary, we may need to refine this a bit.

### DIFF
--- a/R/importWQP.R
+++ b/R/importWQP.R
@@ -58,11 +58,11 @@ importWQP <- function(obs_url, zip=FALSE, tz=""){
       message("zip encoding access still in development")
       temp <- tempfile()
       temp <- paste0(temp,".zip")
-      doc <- GET(obs_url, user_agent(default_ua()), 
-                          write_disk(temp))
-
+      doc <- getWebServiceData(obs_url, write_disk(temp))
       headerInfo <- headers(doc)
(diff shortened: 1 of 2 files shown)


jiwalker-usgs:
fixes #229

R/importWQP.R
```diff
@@ -117,8 +113,6 @@ importWQP <- function(obs_url, zip=FALSE, tz=""){
                                         `HUCEightDigitCode` = col_character()),
                        quote = "", delim = "\t"))
     
-  if(zip) unlink(doc)
```
ldecicco-USGS: Where would this happen now? I think this is a zip file (not via web service)

jiwalker-usgs: I moved it to L65 with an `on.exit`

ldecicco-USGS: Ah OK. When does a zip file over a web service get unzipped?

jiwalker-usgs: It is unzipped L63, that zip is removed L64, and the extracted document is removed after the function returns.

ldecicco-USGS: AAAHHhh, OK, the accordion view had me all messed up. I see it now.

PR #323, opened 2017-02-21
add @importFrom

jiwalker-usgs:
@ldecicco-USGS does this cover it?


ldecicco-USGS:
Then you need to Build...which should update the .Rd file and NAMESPACE

jiwalker-usgs:
Oops, my build config must be messed up, let me fix that.

jiwalker-usgs:
Ok, got my build setup fixed, NAMESPACE changed, Rd did not

PR #326, opened 2017-02-21
Site query

jiwalker-usgs:
This flips the order in request from WQP.  First the parameters from the user are sent to the station service to find out which stations meet the given criteria, then the sites that match are passed into the result service while stripping out the parameters that are only used to determine "spatial" information about the request.  I took a best guess at which parameters these are, and they will have to be maintained with changes to WQP in the future.


R/readWQPdata.R
```diff
@@ -147,35 +147,40 @@ readWQPdata <- function(..., zip=FALSE, querySummary=FALSE){
...
+  values[["mimeType"]] <- "tsv"
+  if (zip) values[["zip"]] <- "yes"
   
-  baseURL <- drURL("wqpData")
-  urlCall <- paste0(baseURL,
-                   urlCall,
-                   "&sorted=no&mimeType=tsv")
-  
-  if(zip) urlCall <- paste0(urlCall,"&zip=yes")
+  urlCall <- drURL("wqpData", arg.list=values)
```
jiwalker-usgs: I switched to this as I needed to build two different urls and didn't want to repeat the above.

jiwalker-usgs:
We may want to hold off on this, the gains seem to be minimal, and we should probably be more discriminating for when we switch to only siteids as it can actually cause it to be quite a bit worse.

ldecicco-USGS:
boo!

Issue #320, opened 2017-02-17
Add countyCd function to readXXX


ldecicco-USGS:
- add ability to lookup countyCd by abbreviation, full name, and code (WQP and maybe NWIS)
- simplify statecode lookup code

PR #327, opened 2017-02-21
add countyCdLookup to readWQPdata and readNWISdata

lindsaycarr:
#320 

WQP does not seem to be returning what I would expect with:

```r
dailyStaffordVA <- readWQPdata(statecode = "Virginia", countycode="Stafford", 
                               parameterCd = "00060")
siteInfo <- attr(dailyStaffordVA, "siteInfo)
unique(siteInfo$countycode) # this should be length 1
```


lindsaycarr:
I keep getting a 400 error, but if I put a breakpoint at the `importWQP` line, it pulls in data. The other weird piece is that the URL constructed within the function correctly has `&countycode=US%3A51%3A011`, but the URL in the message still prints the word. Investigating further tomorrow.

```r
tempAppomattoxVA <- readWQPdata(statecode = "Virginia", countycode="Appomattox", parameterCd = "00010")
Request failed [400]. Retrying in 2 seconds...
Request failed [400]. Retrying in 4 seconds...
For: https://www.waterqualitydata.us/Station/search?statecode=US%3A51&countycode=Appomattox&parameterCd=00010&mimeType=tsv&sorted=no

Error in getWebServiceData(obs_url) : Bad Request (HTTP 400).
```

R/readWQPdata.R
```diff
@@ -114,19 +114,23 @@ readWQPdata <- function(..., zip=FALSE, querySummary=FALSE){
...
-    stCd <- values["stateCd"]
-    if(!grepl("US:",stCd)){
-      values["stateCd"] <- paste0("US:",stateCdLookup(stCd, "id"))
+  names(values)[names(values) == "countyCd"] <- "countycode"
+  if("countycode" %in% names(values)){
+    ctyCd <- values["countycode"]
+    ctyCdPrefix <- paste0(values["statecode"], ":")
+    if(!grepl(ctyCdPrefix, ctyCd)){
+      stCd <- gsub("US:", "", values["statecode"])
+      values["countycode"] <- paste0(ctyCdPrefix, countyCdLookup(stCd, ctyCd, "id"))
```
ldecicco-USGS: I think it would be a lot easier (for me at least) if it was:
```
values["countycode"] <- paste("US",values["statecode"],countyCdLookup(stCd, ctyCd, "id"),sep=":")
```

ldecicco-USGS: and to check whether or not to do that line, add an if statement something like:

```
if(length(grep("US:",values["countycode"])) > 0)){
  values["countycode"] <- paste("US",values["statecode"],countyCdLookup(stCd, ctyCd, "id"),sep=":")
}
```

lindsaycarr: we can't just do `values['statecode']` because that has "US:" attached to it by this time. I can get rid of the prefix object thing though

R/readWQPdots.R
```diff
@@ -18,19 +18,22 @@ readWQPdots <- function(...){
...
-  if("stateCd" %in% names(values)){
-    stCd <- values["stateCd"]
-    if(!grepl("US:",stCd)){
-      values["stateCd"] <- paste0("US:",stateCdLookup(stCd, "id"))
+  names(values)[names(values) == "countyCd"] <- "countycode"
+  if("countycode" %in% names(values)){
+    if(!grepl(values["statecode"], values["countycode"])){
+      stCd <- gsub("US:", "", values["statecode"])
+      values["countycode"] <- paste0(values["statecode"], 
+                                     countyCdLookup(stCd, values["countycode"], "id"))
```
ldecicco-USGS: I'm still confused how the ":" between the state code and county code is getting in there.

lindsaycarr: Oh it was in the previous code with my prefix object, but guess I didn't add it back in appropriately. It didn't seem to make a difference though...

ldecicco-USGS: I'd much rather have it in there, it would match what WQP tells us to include, and is just generally cleaner.

lindsaycarr:
@ldecicco-USGS missing semicolons should no longer magically work

R/readWQPdots.R
```diff
@@ -18,19 +18,23 @@ readWQPdots <- function(...){
...
   
-  if("stateCd" %in% names(values)){
-    stCd <- values["stateCd"]
-    if(!grepl("US:",stCd)){
-      values["stateCd"] <- paste0("US:",stateCdLookup(stCd, "id"))
+  names(values)[names(values) == "countyCd"] <- "countycode"
+  if("countycode" %in% names(values)){
+    if(!grepl(values["statecode"], values["countycode"])){
+      stCd <- gsub("US:", "", values["statecode"])
+      values["countycode"] <- paste(values["statecode"],
```
ldecicco-USGS: Thanks!

PR #328, opened 2017-02-21
Sampling activity

ldecicco-USGS:


tests/testthat/tests_general.R
```diff
@@ -160,6 +160,14 @@ test_that("Dates with no days can be handled", {
   expect_error(readNWISgwl("425957088141001", startDate = "1980-01-01"))
  })
 
+context("whatWQPsamples")
+test_that("whatNWISsites working", {
```
lindsaycarr: whatWQPsamples instead of whatNWISsites here

PR #334, opened 2017-02-24
Improve RDB parser

ldecicco-USGS:


ldecicco-USGS:
This all got on my geojson branch anyway....

PR #335, opened 2017-02-24
Geojson

ldecicco-USGS:


ldecicco-USGS:
Have a look. REAL basic (and kind of neat) test:

```
type <- "Stream"
sites <- whatWQPdata(countycode="US:55:025",siteType=type)
 
library(leaflet)
library(geojsonio)
leaflet() %>% 
     addGeoJSON(geojson_read(attr(sites,"file"))) %>% 
     addProviderTiles("CartoDB.Positron")
```

image

ldecicco-USGS:
OK, so it turns out the only example I have for the bug fix is internal only data (hence...it works on my computer, but fails on Appveyor/Travis). For now, I'm going to comment this test out until I can find another public example.

ldecicco-USGS:
Speak of the devil, I found a case...

FYI...@jiwalker-usgs or @wdwatkins ...travis has been spinning on this all day, but appveyor thinks it's OK. If either of you want to have a look, that would be super.

The thing I think we'll want to debate (either here or in a meeting or something) is if we want `whatWQPdata` to come back with a plain list (like it does right now), and should that list just be a vanilla list, or a valid geojson (which the source is....but that would require bringing in `geojsonio` as a dependency rather than `jsonlite`. `jsonlite` is appealing because it's really basic, streamlined, and has a ton of reverse dependencies. )

Alternatively, we could force it into a data.frame....but frankly I don't see the appeal in that, other than that's what every other `dataRetrieval` function returns.

R/whatWQPdata.R
```diff
@@ -42,29 +39,99 @@ whatWQPsamples <- function(...){
...
+                    "&mimeType=geojson&sorted=no")
+  temp <- tempfile()
+  temp <- paste0(temp,".geojson")
+  doc <- getWebServiceData(urlCall, write_disk(temp))
+  headerInfo <- attr(doc, "headerInfo")
+
+  retval <- fromJSON(temp)
+  attr(retval, "queryTime") <- Sys.time()
+  attr(retval, "url") <- urlCall
+  attr(retval, "file") <- temp
```
jiwalker-usgs: Are we worried about the buildup of temp files here?

ldecicco-USGS: Can you elaborate?

jiwalker-usgs: The response is written to a tempfile here, which is referred to by this attribute.  These will buildup over time as the user makes requests.  I think this is probably fine as it is in the "temp" directory, but I wasn't sure if dataRetrieval had any concern about cleaning these up.

ldecicco-USGS: Yeah. So packages (on CRAN at least) can only save to the temp directory. So, I guess we could add an argument to save somewhere else. But, ideally....I could figure out a way for the list that is made in the R workspace to be ingested into the `geojsonio` workflow (I only spent maybe 5 minutes playing with that)...or....as I keep coming back to, scrape `jsonlite` and read it directly in as geojson with the `geojsonio` package. Another option would be to include in all the examples the step of moving the file to a permanent home (and changing that attribute).

R/getWebServiceData.R
```diff
@@ -52,16 +52,15 @@ getWebServiceData <- function(obs_url, ...){
     
     headerInfo <- headers(returnedList)
 
-    if(headerInfo$`content-type` == "text/tab-separated-values;charset=UTF-8"){
+    if(headerInfo$`content-type` %in% c("text/tab-separated-values;charset=UTF-8")){
```
wdwatkins: What was the motivation for %in%? just curious

ldecicco-USGS: Oh, I had initially put the json content-type in here, but after fiddling realized it needed to be downloaded like the zip....so, it could go back or stay, I don't think it matters

ldecicco-USGS:
I'm going to merge this to get some of the dangling bug fixes in. But, I plan to revisit the return from the geojson function (convert to data frame? keep as vanilla json list? make it official geojson?)

PR #336, opened 2017-02-24
NGWMN tests back in

wdwatkins:


wdwatkins:
The test is failing because the URL is long enough that `getWebService` is trying to POST, which I assume isn't implemented for NGWMN.

Issue #321, opened 2017-02-17
Explore combining USGS & NGWMN water ml2 parser


ldecicco-USGS:

wdwatkins:
There are some differences in the metadata supplied with each time-value pair:
https://cida-test.er.usgs.gov/ngwmn_cache/sos?request=GetObservation&service=SOS&version=2.0.0&observedProperty=urn:ogc:def:property:OGC:GroundWaterLevel&responseFormat=text/xml&featureOfInterest=VW_GWDP_GEOSERVER.USGS.305736084355801

https://waterservices.usgs.gov/nwis/dv/?site=04085427&format=waterml,2.0&ParameterCd=00060&StatCd=00003&startDT=2012-01-01&endDT=2012-06-30

Qualifier/category/value vs comment and source.  It could be doable though to account for both

wdwatkins:
Probably best to keep two functions that read the header, and then call a common `importWaterML2Timeseries` or something function

ldecicco-USGS:
This task is done with #345 
I think we might want to make a separate task to create the logic in `importWaterML2` to handle defaults, a particularly important feature of WaterML2

PR #338, opened 2017-03-03
waterML2 timeseries

wdwatkins:
#321 

Should be passing tests, need to do some testing to make sure other formatting hasn't changed.


R/importNGWMN_wml2.R
```diff
@@ -147,3 +111,68 @@ importNGWMN_wml2 <- function(input, asDateTime=FALSE, tz=""){
...
+  timeDF <- mutate(splitTime, dateTime = NA)
+  logicVec <- nchar(rawTime) > 19
+  if(!all(!logicVec)) { #otherwise sets it to char <NA>
+    timeDF$dateTime[logicVec] <- rawTime[logicVec]
+  }
+  if(asDateTime){
+    timeDF$dateTime <- parse_date_time(timeDF$dateTime, c("%Y","%Y-%m-%d","%Y-%m-%dT%H:%M","%Y-%m-%dT%H:%M:%S",
+                                                          "%Y-%m-%dT%H:%M:%OS","%Y-%m-%dT%H:%M:%OS%z"), exact = TRUE)
+    #^^setting tz in as.POSIXct just sets the attribute, does not convert the time!
+    attr(time, 'tzone') <- tz
```
jiwalker-usgs: Where does `tz` come from?

R/importNGWMN_wml2.R
```diff
@@ -147,3 +111,68 @@ importNGWMN_wml2 <- function(input, asDateTime=FALSE, tz=""){
...
+#' 
+#' Returns data frame columns of all information with each time series measurement;
+#' Anything defined as a default, is returned as an attribute of that data frame.
+#' 
+#' @param input XML with only the wml2:MeasurementTimeseries node and children
+#' @importFrom xml2 xml_attr xml_find_all xml_text 
+#' @importFrom dplyr mutate
+#' @importFrom lubridate parse_date_time
+#' @export
+parseWaterML2Timeseries <- function(input, asDateTime) {
```
wdwatkins: Yeah I was considering that, I kept it here since this is the function that will probably use it the most, since nothing else uses ML2 by default.  I suppose conceptually it fits better with the original though.

R/importNGWMN_wml2.R
wdwatkins: Hmm, looks like no where right now 😬

R/importNGWMN_wml2.R
```diff
@@ -147,3 +111,68 @@ importNGWMN_wml2 <- function(input, asDateTime=FALSE, tz=""){
...
+    timeDF$dateTime <- parse_date_time(timeDF$dateTime, c("%Y","%Y-%m-%d","%Y-%m-%dT%H:%M","%Y-%m-%dT%H:%M:%S",
+                                                          "%Y-%m-%dT%H:%M:%OS","%Y-%m-%dT%H:%M:%OS%z"), exact = TRUE)
+    #^^setting tz in as.POSIXct just sets the attribute, does not convert the time!
+    attr(time, 'tzone') <- tz
+  }
+  
+  uom <- xml_attr(valueNodes, "uom", default = NA)
+  
+  source <- xml_attr(xml_find_all(TVP, ".//wml2:source"), "title")
+  comment <- xml_text(xml_find_all(TVP, ".//wml2:comment"))
```
wdwatkins: NA or empty vector.  I use lines 166-169 to exclude all NA or length 0 columns from the returned data frame.  We can also add/drop things that are more service-specific in the next function up that calls this one.

R/importNGWMN_wml2.R
wdwatkins: will fix and add a test

R/importNGWMN_wml2.R
jiwalker-usgs: I guess my point is whether all of these are the same length so you don't end up with a comment that doesn't match the value.  I believe that this is right, just wasn't sure of the expected return.

wdwatkins: Oh, so if just some some TVPs had comments in the same time series?

wdwatkins: It seems like we should be able to assume that won't happen, but not 100% sure.

R/importNGWMN_wml2.R
wdwatkins: Weird, apparently that hasn't caused an error because there is a `tz` function in lubridate that hides the fact the argument is missing.

R/importNGWMN_wml2.R
```diff
@@ -32,7 +32,7 @@
 #' 
 #' 
 #TODO: separate id and agency name, give also as separate dimensions
-importNGWMN_wml2 <- function(input, asDateTime=FALSE, tz=""){
+importNGWMN_wml2 <- function(input, asDateTime=FALSE, tz){
```
ldecicco-USGS: Any reason not to set the default to "UTC" like the rest of dataRetrivial? I find it confusing when there's a default argument in the middle of a set of arguments, and one that doesn't have a default. Then, I can't just take advantage of that default by saying:
```
importNGWMN(input, "UTC")
```

So, either move `asDateTime` to the right, or add a default of "UTC" to the `tz` (my vote is the 2nd).

wdwatkins: Mm yeah.  I dropped the default since there is already a default coming in from readNGWMNdata, and didn't want to set a default at multiple levels (that was what my question in r-activitites was about).

ldecicco-USGS: gotcha, then take out the default of asDateTime?

ldecicco-USGS:
I'll do that last bit, want to do some tests on my branch with this stuff.

PR #345, opened 2017-03-17
Ngwmn cleanups

ldecicco-USGS:


ldecicco-USGS:
I took out reference to "_wml2" in the names. I think that's not super important to the user. I also tested the timeseries parser on the USGS services. It gets the values and dates, so I think we can just make that our general "base" WaterML2 parser. We should in the future try to get the "default" behavior back in there, and see if we can make it smart enough to handle some comments that are bundled with the TVPs.

wdwatkins:
Yeah, it would be nice to just abstract getting any additional fields within the TVP, if there are additional ones that I didn't code in.

R/importNGWMN_wml2.R
```diff
@@ -7,7 +7,7 @@
 #' @param asDateTime logical, if \code{TRUE} returns date and time as POSIXct, if \code{FALSE}, character
 #' @param tz character to set timezone attribute of datetime. Default is an empty quote, which converts the 
 #' datetimes to UTC (properly accounting for daylight savings times based on the data's provided time zone offset).
-#' Possible values to provide are "America/New_York","America/Chicago", "America/Denver","America/Los_Angeles",
+#' Possible values are "America/New_York","America/Chicago", "America/Denver","America/Los_Angeles",
```
wdwatkins: This is actually anything from `OlsonNames()` now right?

R/importNGWMN_wml2.R
```diff
@@ -112,19 +115,38 @@ importNGWMN_wml2 <- function(input, asDateTime=FALSE, tz){
...
+#' baseURL <- "https://waterservices.usgs.gov/nwis/dv/?format=waterml,2.0"
+#' URL <- paste(baseURL, "sites=01646500",
+#'      "startDT=2014-09-01",
+#'      "endDT=2014-09-08",
+#'      "statCd=00003",
+#'      "parameterCd=00060",sep="&")
+#' \dontrun{
+#' timesereies <- importWaterML2(URL, asDateTime=TRUE, tz="UTC")
+#' } 
+importWaterML2 <- function(input, asDateTime=FALSE, tz="UTC") {
```
wdwatkins: Do want a default tz this far down?  I was going to only include that at the user-interaction level.  And if so, should there be a match.arg for it?  It is currently above in importNGWMN.

R/importNGWMN_wml2.R
ldecicco-USGS: It is...but in the documentation, I want to emphasize that these are the timezones that the US will almost most certainly have. That's why I changed it from "to provide" to "are".

R/importNGWMN_wml2.R
ldecicco-USGS: I left it to be consistent with the other "import" functions. It is possible people use this function as a standalone (the certainly use the other import functions for local files. As WaterML2 becomes even more popular, this function could be used on local files too).

PR #346, opened 2017-03-20
Empty dots

ldecicco-USGS:


tests/testthat/tests_general.R
```diff
@@ -8,6 +8,8 @@ test_that("General NWIS retrievals working", {
   expect_is(multiSite$dateTime, 'POSIXct')
   # saveRDS(multiSite, "rds/multiSite.rds")
   
+  expect_error(readNWISdata())
```
jread-usgs: maybe you want to test for a more specific error here, like:
```r
expect_error(readNWISdata(), "No arguments supplied")
```

R/constructNWISURL.r
```diff
@@ -335,7 +335,7 @@ constructWQPURL <- function(siteNumber,parameterCd,startDate,endDate,zip=FALSE){
 constructUseURL <- function(years,stateCd,countyCd,categories){ 
 
     if(is.null(stateCd)){
-      baseURL <- drURL("useNat", format="rdb", rdb_compression="value")
+      baseURL <- drURL("useNat", format="rdb", rdb_compression="value", Access=pkg.env$access)
```
jread-usgs: seems fine, just wondering if in the future this `Access` param can be added in a single spot for all NWIS calls (maybe right before the `GET`) unless it needs to be added in all spots. Not an issue worth addressing though either way, just a note.

ldecicco-USGS: In the past, adding `Access` to some of the waterdata calls actually broke the links...that's why I was being deliberate. I'll run though all the NWIS services and see if that works now

ldecicco-USGS:
I'm starting to wonder if appveyor has a limit to how much you can do in 1 day? It was chugging along yesterday with my flurry of CI config tests (mainly for travis), then suddenly just stopped working, and now it's back.

Not sure why the coveralls didn't report something. It seems that travis did kick off the coverall script, and it did run. So, I'm going to merge this and do another fairly trivial PR soon. If the coverall doesn't report after that one, I'll keep fiddling with the travis file.

PR #350, opened 2017-03-22
Fixes #317

### LINKED ISSUE #317: biological
Biodata?

ldecicco-USGS:
Show an example of using biodata

### DIFF
--- a/R/readNWISunit.r
+++ b/R/readNWISunit.r
@@ -79,9 +79,8 @@ readNWISuv <- function (siteNumbers,parameterCd,startDate="",endDate="", tz=""){
 #' 
 #' Reads peak flow from NWISweb. Data is retrieved from \url{https://waterdata.usgs.gov/nwis}.
 #' In some cases, the specific date of the peak data is not know. This function will default to
-#' converting the complete dates, dropping rows with incomplete dates. If those incomplete dates are
-#' needed, set the `asDateTime` argument to FALSE. No rows will be removed, and no dates will be converted
-#' to R Date objects.
+#' converting complete dates to a "Date" object, and converting incomplete dates to "NA". If those incomplete dates are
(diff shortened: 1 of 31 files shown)


R/whatWQPdata.R
```diff
@@ -135,6 +135,22 @@ whatWQPdata <- function(..., saveFile = tempfile()){
   
   y[,grep("Count",names(y))] <- sapply(y[,grep("Count",names(y))], as.numeric)
   
+  names(y)[names(y) == "type"] <- paste("type",letters[1:length(names(y)[names(y) == "type"])],sep="_")
```
ldecicco-USGS: @wdwatkins This PR has a ton of changes...but it's mostly just getting the siteNumbers arg consistent, and *not* using siteNumbers as a variable name for examples. The only meat of it is here. I'm checking if the geojson is all "Point"'s, then converting the coordinate list column to regular'ol numeric lat/lon. Jim said currently there's only Points, and no future though of doing anything other than points. So, this should be pretty good for awhile. Also, even if they do add something that's not just a point, this code won't get triggered, so I still think it would be OK. Let me know if you're going to have time to have a look here in the near future, or if you'd like me to find someone else.

R/readNWISunit.r
```diff
@@ -385,11 +384,11 @@ readNWISmeas <- function (siteNumbers,startDate="",endDate="", tz="", expanded=F
 #' @export
 #' @importFrom dplyr left_join
 #' @examples
-#' siteNumber <- "434400121275801"
+#' site_id <- "434400121275801"
 #' \dontrun{
-#' data <- readNWISgwl(siteNumber, '','')
+#' data <- readNWISgwl(site_id, '','')
```
wdwatkins: The `''` arguments are already defaults, but obviously not a big deal.

ldecicco-USGS: True

PR #354, opened 2017-03-27
fix for site with empty timeseries, switch to siteNumbers

wdwatkins:
Not the most elegant, but it works

### DIFF
--- a/R/importNGWMN_wml2.R
+++ b/R/importNGWMN_wml2.R
@@ -84,7 +84,9 @@ importNGWMN <- function(input, asDateTime=FALSE, tz="UTC"){
     }
     nonDateCols <- grep("date",names(mergedDF), value=TRUE, invert = TRUE)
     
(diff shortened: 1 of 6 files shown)


R/importNGWMN_wml2.R
```diff
@@ -146,6 +148,9 @@ importWaterML2 <- function(input, asDateTime=FALSE, tz="UTC") {
   
   gmlID <- xml_attr(returnedDoc,"id") #TODO: make this an attribute
   TVP <- xml_find_all(returnedDoc, ".//wml2:MeasurementTVP")#time-value pairs
+  if(length(TVP) == 0) { #empty nodes on some sites
```
ldecicco-USGS: To follow the pattern of other `dataRetrieval` functions, could you return a data.frame with 0 rows, but the columns that are normally populated? I'm guessing:
```
[1] "source"   "date"     "time"     "dateTime" "value"   
[6] "uom"      "comment" 
```

PR #357, opened 2017-04-05
removed messages, version update for GRAN

wdwatkins:


wdwatkins:
Need to cut a new release.  Can you not delete them?  (I already made one for 2.6.8, but then we fixed up those little things)

PR #375, opened 2017-06-26
Getting ratings in readNWISdata

ldecicco-USGS:


wdwatkins:
That state query should probably work through `readNWISdata` too, right?  It comes in fine with `importRDB1`?

ldecicco-USGS:
The state query does work, I added it to the examples.

Issue #376, opened 2017-06-30
whatNWISdata


ldecicco-USGS:
Incorporate `readNWISdots`, look for other things that might be funky, and add to training material

PR #378, opened 2017-06-30
Cleaning up whatNWISdata

ldecicco-USGS:


ldecicco-USGS:
Seems like the NGWMN tests are failing. Either a hiccup in the service, or something we need to change (I know they were working on some updates....)

@lindsaycarr if you can review sometime next week, the changes in this pr should have no affect on NGWMN...so ignore those errors.

R/whatNWISData.r
```diff
@@ -57,52 +50,48 @@
...
-#' uvData <- whatNWISdata('05114000',service="uv")
-#' uvDataMulti <- whatNWISdata(c('05114000','09423350'),service=c("uv","dv"))
-#' site_ids <- c("01491000","01645000")
-#' flowAndTemp <- whatNWISdata(site_ids, parameterCd=c("00060","00010"))
+#' uvData <- whatNWISdata(siteNumber = '05114000',service="uv")
+#' uvDataMulti <- whatNWISdata(siteNumber = c('05114000','09423350'),service=c("uv","dv"))
+#' flowAndTemp <- whatNWISdata(stateCd = "WI", service = "uv", parameterCd = c("00060","00010"))
 #' }
-whatNWISdata <- function(siteNumbers,service="all",parameterCd="all",statCd="all"){
+whatNWISdata <- function(...){
```
lindsaycarr: I was originally leaving in the previous args so that it wouldn't break people's code that use this function. Think that will be an issue?

R/whatNWISData.r
```diff
@@ -57,52 +50,48 @@
...
-  parameterCds <- unique(SiteFile$parm_cd)
+  values <- sapply(valuesList$values, function(x) URLencode(x))
+
+  if(any(service == "iv")){
+    service[service == "iv"] <- "uv"
+  }
   
-  parameterCdINFO <- parameterCdFile[parameterCdFile$parameter_cd %in% parameterCds,]
-  SiteFile <- merge(SiteFile,parameterCdINFO,by.x="parm_cd" ,by.y="parameter_cd",all=TRUE)
+  if("statCd" %in% names(list(...))){
```
lindsaycarr: instead of testing `names(list(...))`, should it be `matchReturn`?

R/whatNWISData.r
ldecicco-USGS: What will break people's code is if they don't name their arguments now. I'm willing to take that chance for a far more useful function. They could still use:
`siteNumbers,service="all",parameterCd="all",statCd="all"` and that will work (the magic of `dots`)

So really, the script that will break is `whatNWISdata ('01234567')` because it doesn't know that's suppose to be siteNumber, but in exchange we get the full flexibility of dots.

R/whatNWISData.r
ldecicco-USGS: oh yeah, I wrote this before I added that function

R/whatNWISData.r
```diff
@@ -57,57 +50,56 @@
...
+  } else {
+    service <- "all"
   }
   
-  urlSitefile <- drURL('site', Access=pkg.env$access, format='rdb', seriesCatalogOutput='true',sites=siteNumber)
- 
-  SiteFile <- importRDB1(urlSitefile, asDateTime = FALSE)
+  if("statCd" %in% names(matchReturn)){
+    statCd <- matchReturn$statCd
+    matchReturn <- matchReturn[names(matchReturn) != "statCd"]
```
ldecicco-USGS: Leaving this argument in the web call breaks the query

R/whatNWISData.r
```diff
@@ -57,57 +50,56 @@
...
+  urlSitefile <- drURL('site', Access=pkg.env$access, seriesCatalogOutput='true',arg.list=values)
+ 
+  SiteFile <- importRDB1(urlSitefile, asDateTime = FALSE)
+
   if(!("all" %in% service)){
     SiteFile <- SiteFile[SiteFile$data_type_cd %in% service,]
   }
   if(!("all" %in% statCd)){
-    SiteFile <- SiteFile[SiteFile$stat_cd %in% statCd,]
+    SiteFile <- SiteFile[SiteFile$stat_cd %in% c(statCd,NA),]
```
ldecicco-USGS: And this allows for "uv" and "dv" together

R/whatNWISData.r
lindsaycarr: Right, so since you're comfortable then I'm good with it too :) I agree that a more useful function is worth breaking those cases.

lindsaycarr:
Hmm I just pulled down your PR and ran checks that passed with `0 errors | 0 warnings | 0 notes`. I'm going to close and reopen to kick off Travis and Appveyor builds again.

lindsaycarr:
This fixes #376. 

NGWMN must have been experiencing issues last week because this passed locally today, so I restarted the build and now it passes. Might not need to do #379 @wdwatkins? Or look into it but know that it might be hard to recreate.

PR #396, opened 2017-09-27
Having the NGWMN data return even if the site request fails

ldecicco-USGS:


wdwatkins:
Nice, thought we'd have to change more than this.

PR #400, opened 2017-10-27
keep Xs for params that aren't changed

wdwatkins:
If you don't supply the substitution for a column, the column name became all numbers.


wdwatkins:
NGWMN tests are taking forever locally

wdwatkins:
seems like NGWMN tests are working again

PR #407, opened 2017-11-29
fix ML2 parsing, NA removal in ML1

### LINKED ISSUE #392: No data value not turning into NA
https://nwis.waterservices.usgs.gov/nwis/iv/?site=50024950,50027600,50031200,50035000,50038100,50038320,50039500,50043000,50043800,50045010,50046000,50049100,50051800,50055000,50055750,50057000,50106100,50113800,50114900,50136400,50138000,50144000,50147800,50148890&format=waterml,1.1&ParameterCd=00065&startDT=2017-09-17&endDT=2017-09-20

https://github.com/USGS-R/dataRetrieval/blob/master/R/importWaterML1.r#L271

that line is showing FALSE when it needs to be TRUE

wdwatkins:
Apparently the NGWMN functions were hanging due to the parser, not the web service... now fixed.  Still a bit confused about some of the behavior of `xml2::read_xml` when given a nodeset vs a node with children.

also fixes #392

### DIFF
--- a/R/importNGWMN_wml2.R
+++ b/R/importNGWMN_wml2.R
@@ -161,9 +161,9 @@ importWaterML2 <- function(input, asDateTime=FALSE, tz="UTC") {
                       time = character(0), dateTime = character(0), value = numeric(0),
                       uom = character(0), comment = character(0), stringsAsFactors = FALSE))
   }
-  rawTime <- xml_text(xml_find_all(TVP,".//wml2:time"))
+  rawTime <- xml_text(xml_find_all(returnedDoc, "./wml2:point/wml2:MeasurementTVP/wml2:time"))
   
-  valueNodes <- xml_find_all(TVP,".//wml2:value")
+  valueNodes <- xml_find_all(returnedDoc,"./wml2:point/wml2:MeasurementTVP/wml2:value")
   values <- as.numeric(xml_text(valueNodes))
   nVals <- length(values)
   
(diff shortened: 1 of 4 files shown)


R/importNGWMN_wml2.R
```diff
@@ -161,9 +161,9 @@ importWaterML2 <- function(input, asDateTime=FALSE, tz="UTC") {
                       time = character(0), dateTime = character(0), value = numeric(0),
                       uom = character(0), comment = character(0), stringsAsFactors = FALSE))
   }
-  rawTime <- xml_text(xml_find_all(TVP,".//wml2:time"))
+  rawTime <- xml_text(xml_find_all(returnedDoc, "./wml2:point/wml2:MeasurementTVP/wml2:time"))
```
ldecicco-USGS: So why was the old way not working? can we change that xpath on line 158? If not, should we delete that line? What was nice about using the ".//" if I remember correctly was that some WaterML2 had slightly different wrappers than others....(like, feature collections vs wml2 collections vs. single timeseries).

ldecicco-USGS: Did we decide to give up on collections when we moved to `xml2`?

wdwatkins: So, we could still use the old style xpath.  The problem was using `xml_find_all` on `TVP` (a nodeset) instead of `returnedDoc` (a node).  

Since we are searching the tree from higher up now, I thought it would be safer to specify the path in case there was a node with the same name inside the overall metadata section as in the time-value pairs, that could result in a column length mismatch.

wdwatkins: What do you mean by collections?

ldecicco-USGS: When there is more than one site for example and/or more than one paremeter. Take Rhode Island (because it's small):
https://waterservices.usgs.gov/nwis/dv/?format=waterml,2.0&stateCd=ri&siteStatus=all

I think we decided that the WaterML2 parser should be pretty generic....and we were focus on essentially a TVP parser. But I'm not sure I remember exactly

tests/testthat/tests_imports.R
```diff
@@ -114,7 +111,8 @@ test_that("External importWaterML1 test", {
   siteNumber <- c('01480015',"04085427") #one site seems to have lost it's 2nd dd
   obs_url <- constructNWISURL(siteNumber,c("00060","00010"),startDate,endDate,'dv')
   data <- importWaterML1(obs_url)
-  # saveRDS(data, "rds/twoWML1.rds")
+  
+  expect_false(any(data$X_00010_00003) )
```
ldecicco-USGS: Looks like there is mean temperature in the data...what's this test for?

wdwatkins: not sure what is happening here, this is supposed to be this: 
https://github.com/wdwatkins/dataRetrieval/blob/upstreamTest/tests/testthat/tests_imports.R#L179

Checking to see that there aren't any of the no data values left

wdwatkins: hmm, this must have been left from when I started to write this test here and then moved it further down.  I just assumed the error was happening further down at ~#178 :man_facepalming:

PR #410, opened 2017-12-19
different site for no data replace w/NA check

wdwatkins:


tests/testthat/tests_imports.R
```diff
@@ -171,11 +160,11 @@ test_that("External importWaterML1 test", {
 })
 
 test_that("no data val replacement works", {
-  naSite <- "50043800"
-  naData <- readNWISuv(siteNumbers = naSite, parameterCd = "00065",
-                       startDate = "2017-09-17", endDate = "2017-09-20")
-  expect_gt(sum(is.na(naData$X_00065_00000)), 0)
-  expect_false(any(na.omit(naData$X_00065_00000 < -99)))
+  iceSite <- "04024430"
+  naData <- readNWISuv(siteNumbers = iceSite, parameterCd = "00060",
```
ldecicco-USGS: might want to put a comment in here that generally when the hydrologists start reviewing their data...the NA's go away because they have ways to estimate the value (so probably every year this test needs to be updated to the latest winter).

wdwatkins: Yeah, is it worth testing this if it has to be changed constantly?

ldecicco-USGS: That's why I commented it out initially (looks like 3 year-ish ago). I guess I'd vote to comment the test out (leave the commented-out code here though), and add a note by the test that the dates will most likely need to be updated to get NA/Ice values.

If we mess with anything that affects the possible output of NA values...we need to remember to run that test. It's not ideal, but probably will cause less headaches than leaving it in?

wdwatkins: Yeah I suppose so.  Looks like they eventually remove the iced out parts:
image

PR #411, opened 2018-01-02
Some comments start with quotes, but don't end with them...

ldecicco-USGS:
Thanks Excel!


ldecicco-USGS:
Here is an example that wasn't working:
```
well87<-425651070573701

ope87<-readNWISqw(well87,"OPE","","")

```

PR #412, opened 2018-01-03
fix to NGWMN code and tests

wdwatkins:


R/importNGWMN_wml2.R
```diff
@@ -207,7 +210,7 @@ importWaterML2 <- function(input, asDateTime=FALSE, tz="UTC") {
                   uom = uom, comment = comment)
   df_use <- df_vars[sapply(df_vars, function(x){length(x) > 0 && !all(is.na(x))})]
   df <- data.frame(df_use, stringsAsFactors = FALSE)
-  
+  if(!"value" %in% names(df)) {df$value <- NA}
```
ldecicco-USGS: I think you should make this:
```
df$value <- as.numeric(NA)
```

tests/testthat/tests_imports.R
```diff
@@ -188,7 +188,7 @@ test_that("importWaterML2 external test", {
   xml <- getWebServiceData(url)
   measurementTS2 <- xml2::xml_find_all(xml, "//wml2:MeasurementTimeseries")
   exData <- importWaterML2(measurementTS2)
-  expect_is(exData$value, 'numeric')
+  expect_is(exData$value, c('numeric', 'logical'))
```
ldecicco-USGS: Presumably, this test will go back to all numeric's as soon as the ice melts. Would it be a more stable test if it was:
```
expect_true(class(exData$value) %in% c('numeric', 'logical'))
```
I'm actually still a bit confused why it would come out as logical...even if they are all NA's...I'd think we'd want it numeric NA's...found the problem line I think above.

PR #463, opened 2019-04-03
Checking if travis can run examples

ldecicco-USGS:
Added `--run-donttest`....it might take more than 60 minutes though....we'll see


ldecicco-USGS:
I changed all the "dontrun" to "donttest". This required a bunch of small fiddling (like, commenting out lines that seemed commented out in the roxygen part of the example code) to get travis back to passing. It takes A LOT longer to check now, but "donttest" is the preferred way on CRAN. They won't test the examples...but we still can. "dontrun" is used to indicate "really...don't run this!"

Anyway, I'm going to merge this so I can have a separate pull request that includes some actual changes to the code

PR #469, opened 2019-04-10
Updating github site

ldecicco-USGS:


ldecicco-USGS:
Mostly all updates to the help files that would show up here:
http://usgs-r.github.io/dataRetrieval/
and one tiny fix

Issue #434, opened 2018-08-09
use POST instead of GET for downloading data from the Water Quality Portal


jkreft-usgs:
There are a variety of ways that GETs can end up breaking, such as URLS being to long, problems escaping characters, to firewalls interpreting queries as attacks and blocking them, to apache routing getting all confused.   Many of these problems can be avoided by POSTing a JSON payload instead.  

The syntax is pretty reasonable, and is documented in the  [swaggerdocs](https://www.waterqualitydata.us/data/swagger-ui.html#!/Result/resultJsonPostRequestUsingPOST) and also below.  
you do need to send the headers, and include the mimetype in the URL.  here is an example curl:
`curl -X POST --header 'Content-Type: application/json' --header 'Accept: application/zip' -d '{"statecode":["US:55"],"countycode":["US:55:025","US:55:049"],"siteType":["Lake, Reservoir, Impoundment","Stream"],"sampleMedia":["Air","Water"]}' 'https://www.waterqualitydata.us/data/Station/search?mimeType=csv&zip=yes'`

```
{
  "analyticalmethod": [
    "string"
  ],
  "assemblage": [
    "string"
  ],
  "bBox": "string",
  "characteristicName": [
    "string"
  ],
  "characteristicType": [
    "string"
  ],
  "countrycode": [
    "string"
  ],
  "countycode": [
    "string"
  ],
  "dataProfile": "string",
  "huc": [
    "string"
  ],
  "lat": "string",
  "long": "string",
  "mimeType": "string",
  "minactivities": "string",
  "minresults": "string",
  "nldiurl": "string",
  "organization": [
    "string"
  ],
  "pCode": [
    "string"
  ],
  "project": [
    "string"
  ],
  "providers": [
    "string"
  ],
  "sampleMedia": [
    "string"
  ],
  "siteType": [
    "string"
  ],
  "siteid": [
    "string"
  ],
  "startDateHi": "string",
  "startDateLo": "string",
  "statecode": [
    "string"
  ],
  "subjectTaxonomicName": [
    "string"
  ],
  "within": "string",
  "zip": "string"
}
```

ldecicco-USGS:
See @limnoliver  issue:
```r
readWQPdata(siteid = 'COE/ISU_WQX-17630001', characteristicName = 'Temperature, water')
```

ldecicco-USGS:
What I've learned....
[...]

PR #475, opened 2019-04-24
Changing WQP to default to "zip"

ldecicco-USGS:
Significant time improvements...should have been done awhile ago. I want to get this PR through before trying the "POST" updates.


R/getWebServiceData.R
```diff
@@ -22,7 +22,11 @@
 #' }
 getWebServiceData <- function(obs_url, ...){
   
-  returnedList <- retryGetOrPost(obs_url, ...)
+  if(grepl(pattern = "waterqualitydata", x = obs_url)){
```
ldecicco-USGS: FYI... this is kind of paving the way for defaulting to all POSTs in WQP

R/importWQP.R
```diff
@@ -116,29 +117,60 @@ importWQP <- function(obs_url, zip=FALSE, tz="UTC"){
...
   
+}
+
+
+post_url <- function(obs_url, zip){
+
+  split <- strsplit(obs_url, "?", fixed=TRUE)
+  
+  url <- split[[1]][1]
+  url <- paste0(url, "?mimeType=tsv")
```
wdwatkins: Does this function need the csv/tsv argument passed in?

R/readNGWMNdata.R
```diff
@@ -119,19 +118,19 @@ readNGWMNdata <- function(service, ..., asDateTime = TRUE, tz = "UTC"){
 #' \donttest{
 #' #one site
 #' site <- "USGS.430427089284901"
-#' oneSite <- readNGWMNlevels(siteNumbers = site)
+#' #oneSite <- readNGWMNlevels(siteNumbers = site)
 #' 
 #' #multiple sites
 #' sites <- c("USGS:272838082142201","USGS:404159100494601", "USGS:401216080362703")
-#' multiSiteData <- readNGWMNlevels(sites)
```
wdwatkins: are these not working again?

vignettes/usMaps.Rmd
```diff
@@ -74,6 +76,8 @@ shift_sp <- function(sp, scale, shift,
 
 ```
 
+For the United States, this function will get call the `shift_sp` function on the state/territory boundaries.
```
wdwatkins: Don't need to deal with it now, but at some point it might be good to switch this to `sf`.  Could be a good exercise for someone looking to learn `sf`?

ldecicco-USGS: Yeah, I wanted to get this out because it was so helpful for the gage outage map. But, if someone looked at it and said, I could re-do that in `sf`, that would be great. In fact, someone at an unConf has already started I think

R/getWebServiceData.R
```diff
@@ -106,6 +107,29 @@ getQuerySummary <- function(url){
   return(retquery)
 }
 
+retryPOST <- function(obs_url, ...) {
```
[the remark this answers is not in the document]
ldecicco-USGS: If we do move to a `POST`, there are enough things unique from NWIS, that I made it it's own special function so when we do start that work, we are certainly not breaking anything on NWIS calls.

R/getWebServiceData.R
```diff
@@ -106,6 +107,29 @@ getQuerySummary <- function(url){
...
+
+  if (nchar(obs_url) < 2048){
+    resp <- httr::RETRY("GET", 
+                      obs_url, ..., 
+                      httr::user_agent(default_ua())) 
+  } else {
+    split <- strsplit(obs_url, "?", fixed=TRUE)
+    obs_url <- split[[1]][1]
+    query <- split[[1]][2]
+    # resp <- httr::RETRY("POST",
```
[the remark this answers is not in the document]
ldecicco-USGS: Yeah, it took me a little bit of futzing to figure out the `httr::content_type("application/json")` which is why I left it commented in there (in case I wasn't the one who would work on it down the road). But...I'll put a few notes in #434

R/getWebServiceData.R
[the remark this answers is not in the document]
wdwatkins: should it be `retryWQPPOST` or something then?

R/readNGWMNdata.R
ldecicco-USGS: They are taking forever....I saw in slack they were having some hiccups again.

wdwatkins: Maybe (in the future) it would be worth building in a `verbose` argument that if true tells you when different steps of the call finish?  I had built in at one point a message when the file was downloaded so you knew when something taking forever was due to the download time or parsing issues.

ldecicco-USGS: there's a verbose option via httr...but yeah, we could add on to that. Want to make an issue?

wdwatkins: sure

R/whatWQPdata.R
```diff
@@ -22,7 +22,7 @@ whatWQPsamples <- function(...){
                     urlCall,
                     "&mimeType=tsv")
   
-  retval <- importWQP(urlCall, zip=values["zip"] == "yes")
+  retval <- suppressWarnings(importWQP(urlCall, zip=values["zip"] == "yes"))
```
wdwatkins: I think there was something in particular here generating an innocuous warning, right?  Is there any way you could more surgically handle that warning without suppressing others?

ldecicco-USGS: yeah, every R update or so I cross my fingers that they'll just add that option to suppressWarning for a specific regex.... But, I guess the language translation issue would make that impossible.  `suppressWarning` is one of those things I use the lazy solution because the `withCallingHandlers`+`invokeRestart` makes me more nervous that I'll screw it up.  The other option of course is to leave it in. Jim says they will fix it eventually....

wdwatkins: I'd say leave the warning, or make an issue to document why the `suppressWarning` is there/when it can be removed.

PR #487, opened 2019-08-20
Update the code.json file

ldecicco-USGS:
This is a demo of github, but it has an update to the code.json file.

### DIFF
--- a/code.json
+++ b/code.json
@@ -3,7 +3,7 @@
     "name": "dataRetrieval",
     "organization": "U.S. Geological Survey",
     "description": "This R package is designed to obtain USGS or EPA water quality sample data, streamflow data, and metadata directly from web services",
-    "version": "2.7.4",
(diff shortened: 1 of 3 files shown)


code.json
```diff
@@ -17,7 +17,7 @@
     },
 
     "homepageURL": "https://code.usgs.gov/WMA-Data-Science/dataRetrieval",
-    "downloadURL": "https://code.usgs.gov/WMA-Data-Science/dataRetrieval/-/archive/v2.7.1/dataRetrieval-v2.7.1.zip",
+    "downloadURL": "https://code.usgs.gov/WMA-Data-Science/dataRetrieval/-/archive/v2.7.5/dataRetrieval-v2.7.5.zip",
```
ldecicco-USGS: The version was updated, is this correct in code.usgs.gov

PR #535, opened 2020-10-16
avoid partial matching of headerInfo, closes #534

### LINKED ISSUE #534: Avoid partial matching on the headerInfo attribute of webservice objects
I get a warning on running the first example for `readNWISqw`:

> readNWISqw
In attr(doc, "header") : partial match of 'header' to 'headerInfo'

Under the hood `getWebServiceData.r` sets an attribute `headerInfo` but subsequent lines in `importRDB1.r` and `readNWISqw.r` read from the `header` attribute (instead of `headerInfo`) via partial matching. While partial matching doesn't cause an error in this case, it might in the future. I have a PR prepared to read explicitly from `headerInfo` and avoid this warning.

You may not see the same warning unless you have the following set in your `.Rprofile`:
> warnPartialMatchAttr = TRUE

jsta:
See #534

### DIFF
--- a/R/importRDB1.r
+++ b/R/importRDB1.r
@@ -91,9 +91,9 @@ importRDB1 <- function(obs_url, asDateTime=TRUE, convertType = TRUE, tz="UTC"){
     doc <- obs_url
   } else {
     doc <- getWebServiceData(obs_url, encoding='gzip')
-    if("warn" %in% names(attr(doc,"header"))){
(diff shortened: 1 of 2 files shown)


ldecicco-USGS:
Thanks for the PR! Just clicked your user name and see you are in Madison now, cool! If there wasn't a pandemic, I'd say we should have a meet and greet with other USGS-R people. Maybe summer 2021.... 😭

PR #571, opened 2021-05-13
Avoid operating on the readr problems attribute directly

jimhester:
In the upcoming readr 2.0.0 release the internals of the problems
attribute have changed. It is no longer a tibble of problems, instead it
is an external pointer to the index that is converted to a tibble by the
`problems()` function.

The public interface to retrieve the problems was always using the
`problems()` function, these code changes in this commit convert the
existing code to use only the public interface.

The current code will cause tests to fail when readr 2.0.0 is on CRAN so
this change should be submitted before that release to avoid failures on
CRAN.

I verified that these changes fix the CRAN test failures, but I didn't run 
or try to fix any potential  non-CRAN test failures.

### DIFF
--- a/R/importRDB1.r
+++ b/R/importRDB1.r
@@ -146,7 +146,8 @@ importRDB1 <- function(obs_url, asDateTime=TRUE, convertType = TRUE, tz="UTC"){
       char.names <- NULL
     } 
     
-    if(nrow(readr::problems(readr.data)) > 0 | length(char.names) > 0){
+    readr.problems <- readr::problems(readr.data)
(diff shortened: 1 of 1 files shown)


ldecicco-USGS:
Thanks @jimhester ! I'm working on a PR soon that cleans up even more of the readr-early-adapter-hacks I was doing. I'll try to get an update to CRAN soon. Do you have an estimate on when readr 2.0's going to be submitted?

jimhester:
I would guess 3-5 weeks from now, if all goes well.

PR #581, opened 2021-09-10
Failing gracefully

ldecicco-USGS:
We've been pinged for not failing gracefully on our examples in CRAN (this is the 2nd time). CRAN doesn't want the functions to `stop` or `error` if the server returns an error message. So...I switched the `getWebService` function to use `httr::message_for_status` (instead of `httr::stop_for_status`), and return a `NULL`. This is consistent with what @dblodgett-usgs  is doing with `sbtools`.

@wdwatkins or @dblodgett-usgs ...if one of you could glance at the code to see if it looks good to you, that would be great. I actually got to test it quite extensively yesterday because WQP was failing all over the place (someone was overloading them).

We need to resubmit to CRAN in the next couple of days to avoid having it taken down.

### DIFF
--- a/R/importNGWMN_wml2.R
+++ b/R/importNGWMN_wml2.R
@@ -25,9 +25,9 @@
 #' "observedProperty=urn:ogc:def:property:OGC:GroundWaterLevel",
 #' "responseFormat=text/xml",
 #' "featureOfInterest=VW_GWDP_GEOSERVER.USGS.403836085374401",sep="&")
-#' if(!httr::http_error(obs_url)){
-#'   data_returned <- importNGWMN(obs_url)
-#' }
+#' 
+#'  data_returned <- importNGWMN(obs_url)
+#' 
 #' }
 #' 
 importNGWMN <- function(input, asDateTime=FALSE, tz="UTC"){
--- a/R/importNGWMN_wml2.R
+++ b/R/importNGWMN_wml2.R
@@ -44,7 +44,9 @@ importNGWMN <- function(input, asDateTime=FALSE, tz="UTC"){
     raw <- TRUE
   } else {
     returnedDoc <- getWebServiceData(input, encoding='gzip')
-    
+    if(is.null(returnedDoc)){
+      return(invisible(NULL))
+    }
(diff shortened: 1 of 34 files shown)


R/importNGWMN_wml2.R
```diff
@@ -43,7 +44,9 @@ importNGWMN <- function(input, asDateTime=FALSE, tz="UTC"){
     raw <- TRUE
   } else {
     returnedDoc <- getWebServiceData(input, encoding='gzip')
-    
+    if(is.null(returnedDoc)){
```
ldecicco-USGS, line 47: Since the function doesn't stop in the `getWebServiceData`, we now have to add these all over the place

R/importNGWMN_wml2.R
```diff
@@ -25,8 +25,9 @@
 #' "observedProperty=urn:ogc:def:property:OGC:GroundWaterLevel",
 #' "responseFormat=text/xml",
 #' "featureOfInterest=VW_GWDP_GEOSERVER.USGS.403836085374401",sep="&")
-#' data <- importNGWMN(obs_url)
-#' 
+#' if(!httr::http_error(obs_url)){
+#'   data_returned <- importNGWMN(obs_url)
+#' }
```
dblodgett-usgs, line 30: Does this hit that URL twice?

R/importNGWMN_wml2.R
dblodgett-usgs: The other way to handle would be a trycatch with a message and return NULL in the catch.

dblodgett-usgs:
This seems reasonable -- Thinking about how it would have been done if this "no errors ever on web service calls" had been the initial design, I feel like it would have ultimately done what you are doing if in a little less "bolt on" way. I'd say :shipit:

R/importNGWMN_wml2.R
ldecicco-USGS: This was the original issue because the cida link was 502'ing. I added this check because it's the sketchiest source. I could take that out though...

ldecicco-USGS:
Daw...I haven't seen the ship-it squirrel in forever!

PR #584, opened 2021-09-24
quote handling -- fixes #583

### LINKED ISSUE #583: Add quote argument test
This site:
https://waterservices.usgs.gov/nwis/site/?siteOutput=Expanded&format=rdb&site=11486500
has a double quote in the name

dblodgett-usgs:
I found a site in Wisconsin as noted in #583 with the name `"FISHER CR AT 32 & HIGHLAND RD AT HOWARDS GROVE, W`

That was breaking the readr parsing without specifying no quotes. I've run this over the whole site file and it seems to work fine. I would not be surprised to find out that there _are_ quotes used as string quotes some place, but the tab separator seems to preclude the need for them? Unless there are tabs in the strings too -- Someone totally but a \t in a string in nwis some place...

### DIFF
--- a/R/findNLDI.R
+++ b/R/findNLDI.R
@@ -23,7 +23,8 @@ find_good_names = function(input, type) {
   if (type == "nav") {
     grep("comid", names(input), value = TRUE)
   } else if (type == "feature") {
-    c("sourceName", "identifier", "comid")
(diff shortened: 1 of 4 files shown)


dblodgett-usgs:
Now also fixes #587

Issue #591, opened 2021-12-08
readNWISpCode("all") not parsing as expected


lindsayplatt:
**Describe the bug**
`readNWISpCode("all")` is no longer working. No longer returns a nicely parsed dataset.

**To Reproduce**
I am using version 2.7.10 (but colleagues experienced this on 2.7.1). When you run `readNWISpCode("all")`, you do not get a nicely parsed table, but instead the start of an HTML webpage:

```                                                                                                                                                                             NA.
1                                                                                                                                                                                                                                                 <html xmlns="http://www.w3.org/1999/xhtml" lang="en">
2                                                                                                                                                                                                                                                                                                <head>
3                                                                                                                                                                                                                                 <meta http-equiv="Content-Type" content="text/html; charset=utf-8" />
4                                                                                                                                                                                        <base href="https://help.waterdata.usgs.gov/codes-and-parameters/codes" /><!--[if lt IE 7]></base><![endif]-->
5         <meta content="Search criteria are criteria that you enter to select sites of interest. Codes describe data and aid in its interpretation. This page contains links to a comprehensive set of codes used by this site which can be used an an authoritative reference." name="description" />
6                                                                                                                       <link rel="stylesheet" type="text/css" media="screen" href="https://help.waterdata.usgs.gov/portal_css/Sunburst%20Theme/reset-cachekey-8d2e968c3efbf6373cda549130a0c279.css" />
```


**Expected behavior**
I expect to see a table akin to what is returned from `parameterCdFile`. I instructed colleagues to use `parameterCdFile` instead for now to the get all the pcode info into a table.

**Session Info**
Using `dataRetrieval` version 2.7.10.
```r
R version 4.1.1 (2021-08-10)
Platform: x86_64-w64-mingw32/x64 (64-bit)
Running under: Windows 10 x64 (build 19042)
```

**Additional context**
Colleagues first told me about this failing yesterday, but I don't know when it worked last to say that yesterday was the first day. Could it be related to the Drupal 9 switch?

ldecicco-USGS:
Ugg...yeah, that's new. It also doesn't work when the user asks for a pcode that's not in the parameterCdFile. Usually, that would kick off a call to the service. Now you get a similar html result.

bhuffman-usgs:
[...]

PR #593, opened 2021-12-20
Parameter code lookup

ldecicco-USGS:
Old service is now returning html


ldecicco-USGS:
Fixes #591
Thanks @bhuffman-usgs

Issue #623, opened 2022-07-06
findNLDI returns error when API return empty json


mps9506:
Hi,

This isn't fully a bug. 

Currently `findNLDI` returns a somewhat cryptic message and errors when making a query that returns nothing (valid query but nothing is found). 

 For example I know the following will return nothing:
 https://labs.waterdata.usgs.gov/api/nldi/linked-data/wqp/TCEQMAIN-10016/navigation/UM/nwissite?f=json&distance=2
```
{"type":"FeatureCollection"}
```

Here is what `findNLDI` returns when it encounters the same thing:
```r
findNLDI(wqp = "TCEQMAIN-10016",
    nav = "UM",
    find = "nwissite",
    distance_km = 2,
    no_sf = FALSE)

Error: Cannot open "{"type":"FeatureCollection"}"; Check connection parameters.
In addition: Warning messages:
1: In CPL_read_ogr(dsn, layer, query, as.character(options), quiet,  :
  GDAL Error 1: Invalid FeatureCollection object. Missing 'features' member.
2: In CPL_read_ogr(dsn, layer, query, as.character(options), quiet,  :
  GDAL Error 1: Layer schema generation failed.
3: In CPL_read_ogr(dsn, layer, query, as.character(options), quiet,  :
  GDAL Error 4: Failed to read GeoJSON data
```

I think for a user expecting a response, the current error and warning messages are hard to troubleshoot.  The previous iteration of  `navigate_nldi` in nhdPlusTools returned an empty tibble in a similar scenario: https://github.com/USGS-R/nhdplusTools/issues/97.  Having a message indicating that no features were returned and returning a NULL or empty object/list might be beneficial for users.

As always, thanks for this awesome package!!! Happy to make a pull request if there is a way you prefer this handled.

ksonda:
Possibly related https://github.com/internetofwater/nldi-services/issues/315

ldecicco-USGS:
@dblodgett-usgs or @mikejohnson51 can one of you take a look?

mps9506:
@mikejohnson51 
Thanks! I think this approach plus a message makes sense.

mikejohnson51:
Awesome - now were looking at something like this:

``` r
library(dataRetrieval)
xx = findNLDI(wqp = "TCEQMAIN-10016",
              nav = "UM",
              find = "nwissite",
              distance_km = 2,
              no_sf = FALSE)
#> No data found for: nwissite?distance=2

xx
[...]

PR #624, opened 2022-07-08
Adapting NLDI client to changes in the NLDI service

mikejohnson51:
Hi @ldecicco-USGS,

This PR modifies the `findNLDI` function to handle two changes in the NLDI response. 

The first was raised in #623 and relates to requests that return no data. The behavior now is a more clear warning message that no data was found rather then the crypitc `jsonlite`/`sf` error message.

The second was a change in the NLDI that defaulted to HTML landing pages (raised in internetofwater/nldi-services#315). The behavior now is that a JSON object is explicitly requested in the URLs constructed in `findNLDI` (opposed to relying on a default).

All tests pass locally as well. Please let me know if anything else needs changing!

Mike

### DIFF
--- a/R/findNLDI.R
+++ b/R/findNLDI.R
@@ -89,28 +89,41 @@ get_nldi = function(url, type = "", use_sf = FALSE) {
     # Interpret as text
     d <- httr::content(res, "text", encoding = "UTF8")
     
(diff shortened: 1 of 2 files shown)


ldecicco-USGS:
Awesome, looks great!

PR #626, opened 2022-07-14
Work with 0 length responses more appropriatly in #623 

dblodgett-usgs:


dblodgett-usgs:
OK @ldecicco-USGS and @mikejohnson51 -- this should be good to go now.

PR #627, opened 2022-08-03
Initial bug fix work

### LINKED ISSUE #619: readNWISdata with peak service
I would expect this to work:
```r
peaks <- readNWISdata(stateCd="PA", service="peak")
```
But, you need this instead:
```r
peaks <- readNWISdata(stateCd="PA", service="peak",
                      list_of_search_criteria="state_cd")
```

We should check state, county, and HUC queries and make sure the peak requests act similar to the other services.

alkrall:
Fixes #619

### DIFF
--- a/tests/testthat/tests_general.R
+++ b/tests/testthat/tests_general.R
@@ -106,7 +106,14 @@ test_that("General NWIS retrievals working", {
   expect_equal(2, nrow(multi_huc))
 
   
-  
+ peak_data <- readNWISdata(service = "peak", 
+                           state_cd = "PA")
+ expect_lt(nrow(peak_data, 100000))
+ 
+ peak_data <- readNWISdata(service = "peak", 
+                           huc2_cd = "20")
+ expect_lt(nrow(peak_data, 100000))
+ 
 })
 
 test_that("whatNWISdata",{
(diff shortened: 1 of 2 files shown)


tests/testthat/tests_general.R
```diff
@@ -106,7 +106,14 @@ test_that("General NWIS retrievals working", {
   expect_equal(2, nrow(multi_huc))
 
   
-  
+ peak_data <- readNWISdata(service = "peak", 
+                           state_cd = "PA")
+ expect_lt(nrow(peak_data, 100000))
```
ldecicco-USGS, line 111: I think this needs to be:
```r
expect_lt(nrow(peak_data), 100000)
```
(same with down below)

ldecicco-USGS:
Try updating the tests:
```r
expect_lt(nrow(peak_data), 100000)
```
(so the parenthesis is moved), and push that change up. The tests will all kick off again when you click Push. Before you do that, make sure you change your branch back to "main" (it's in that dropdown in the upper right)

ldecicco-USGS:
I think this is all good now. There are some new NOTES popping up, but I think those are new because of a newer devel R version. Also need to check with GH Actions if there's a better way to get sf to install.

Issue #610, opened 2022-02-17
Add convertType to WQP functions


ldecicco-USGS:
Just like the NWIS functions, there should be an argument in the WQP functions to bring in everything as character instead of having dataRetrieval try to make the correct dates/numerics/etc.

alkrall:
1) Added: 
```r
convertType=TRUE
```
if statement to `importWQP`.
2) Added convertType to function call in `importWQP`, `readWQPdata`, `readWQPqw`, `whatWQPdata`.
3) Set: 
```r
convertType=convertType 
```
to `importWQP` function call within `readWQPdata`, `readWQPqw`, and `whatWQPdata`. 
4) Added reoxygen documentation to `importWQP`, `readWQPdat`a, `readWQPqw`, `whatWQPdata`.
5) Added tests to testthat.R.
	-Test issues raised in `whatWQPsamples` and `whatWQPmetrics` with `importWQP`.

PR #632, opened 2022-08-16
Fixes to issue #610.

alkrall:
Fixes to issue #610


R/importWQP.R
```diff
@@ -145,6 +151,9 @@ importWQP <- function(obs_url, zip=TRUE, tz="UTC",
       attr(emptyReturn$ActivityStartDateTime, "tzone") <- tz
       attr(emptyReturn$ActivityEndDateTime, "tzone") <- tz
       
+      if(!convertType){
```
ldecicco-USGS, line 154: I guess if convertType is true, we would want to make this empty data frame completely character

R/whatWQPdata.R
```diff
@@ -31,7 +31,7 @@ whatWQPsamples <- function(...){
   baseURL <- appendDrURL(baseURL, mimeType = "tsv")
   
   withCallingHandlers({
-    retval <- importWQP(baseURL, zip=values["zip"] == "yes")
+    retval <- importWQP(baseURL, zip=values["zip"] == "yes", convertType=convertType)
```
ldecicco-USGS, line 34: I don't know where this convertType value would come from. I think we want to add it to the function itself (?). So something like this:
```r
whatWQPsamples <- function(..., convertType = TRUE){

```
Then you'd need to add the @param stuff to this function.

R/whatWQPdata.R
```diff
@@ -74,7 +76,7 @@ whatWQPmetrics <- function(...){
...
```
ldecicco-USGS, line 79: same as whatWQPsamples above

ldecicco-USGS:
When you click "Install" from the Build tab, you should be updating the .Rd files. If not, in that same Build tab, click More -> Configure Build Tools -> then click the checkbox for "Generate documentation with Roxygen". 

Once that's done, you should see a few .Rd files that need to be committed and then Pushed up to this PR. The R Check that is giving some Warnings are saying the help files need some information on this new argument "convertType". I see you updated the Roxygen code above the functions themselves, but we need that Roxygen code to actually update the .Rd files in the man folder.

R/importWQP.R
alkrall: Updated all columns to character data type within the  'emptyReturn' data frame.

alkrall:
Hi Laura,
Let me know if I addressed these correctly.

Aliesha L. Krall
Physical Scientist
U.S. Geological Survey Upper Midwest Water Science Center
2280 Woodale Drive
Mounds View, MN 55112
office: 763-783-3135
https://www.usgs.gov/staff-profiles/aliesha-l-diekoff
________________________________
From: Laura DeCicco ***@***.***>
Sent: Monday, August 22, 2022 1:12 PM
To: USGS-R/dataRetrieval ***@***.***>
Cc: Krall, Aliesha L ***@***.***>; Author ***@***.***>
Subject: [EXTERNAL] Re: [USGS-R/dataRetrieval] Fixes to issue #610. (PR #632)




 This email has been received from outside of DOI - Use caution before clicking on links, opening attachments, or responding.



When you click "Install" from the Build tab, you should be updating the .Rd files. If not, in that same Build tab, click More -> Configure Build Tools -> then click the checkbox for "Generate documentation with Roxygen".

Once that's done, you should see a few .Rd files that need to be committed and then Pushed up to this PR. The R Check that is giving some Warnings are saying the help files need some information on this new argument "convertType". I see you updated the Roxygen code above the functions themselves, but we need that Roxygen code to actually update the .Rd files in the man folder.

R/importWQP.R
ldecicco-USGS: We don't want to make everything character unless the user choose convertType = FALSE. So we'll need to get the emptyReturn back to it's original types (numerics, dates, etc).

Then, you can add something like:
```r
if(!convertType){
  i <- sapply(emptyReturn, !is.character)
  emptyReturn[i] <- lapply(emptyReturn[i], as.character)
}
```
(you'll need to double-check that code....)

tests/testthat/tests_general.R
```diff
@@ -556,3 +556,21 @@ test_that("readWQPsummary", {
...
+  
+  pH <- readWQPdata(statecode="WI", countycode="Dane", characteristicName="pH", convertType=FALSE)
+  expect_is(pH$ResultMeasureValue, "character")
+  
+  SC<-readWQPqw(siteNumbers="USGS-05288705", parameterCd="00300", convertType=FALSE)
+  expect_is(SC$ResultMeasureValue, "character")
+  
+  lakeSites_chars <- whatWQPdata(siteType = "Lake, Reservoir, Impoundment", 
+                                 statecode = "US:55", convertType=FALSE)
+  expect_is(lakeSites_chars$lat, "character")
```
ldecicco-USGS, line 574: All the github tests are failing here, they're showing that the lat column is coming in as numeric. Could you check on your end if you also get a numeric?

ldecicco-USGS:
We're going pull this, and either myself or Aliesha will fix the broken tests.

Issue #682, opened 2023-11-21
broken web services link in whatNWISdata documentation


lstanish-usgs:
**Describe the bug**
The waterservices documentation was recently updated, and it appears that the URLs have changed. I found one example in the documentation for ?whatNWISdata, which is now broken. The correct URL for the site service is now: https://waterservices.usgs.gov/docs/site-service/site-service-details/.
One option is to dial back the level of resolution in the URL to just waterservices.usgs.gov, which should be stable. The documentation could direct users to the specific service without including it in the URL link.

**To Reproduce**
```r
library(dataRetrieval)
?whatNWISdata
```r
click on waterservices link

**Expected behavior**
The site service info page should load. Correct URL: https://waterservices.usgs.gov/docs/site-service/site-service-details/

PR #683, opened 2023-11-29
Issue682: fix waterservice links and screenshots

rckwzrd:
#682

Links to water service test tools and documentation have changed. Links to water services have been updated in R/ and vignettes/. New water service screenshot also provided. Links in man/ not touched. Assuming that documentation files in man/ are generated from R/ with roxygen. 

There are still references to old water services links in data output files in inst/extdata but I don't think those need to be updated.

No changes to actual functions, logic, or R code itself.


lstanish-usgs:
@ldecicco-USGS does this catch all the requisite updates?

ldecicco-USGS:
Yep! Looks great!

PR #696, opened 2024-02-20
Feb -> March

ldecicco-USGS:
Also a lot of troubleshooting on a GitLab CI job failing. Works now!

### DIFF
--- a/vignettes/Status.Rmd
+++ b/vignettes/Status.Rmd
@@ -21,7 +21,7 @@ This page will be updated frequently with information about the status of dataRe
 
 ### NWIS qw services
 
-Discrete water samples data are undergoing modernization, and NWIS services will no longer be updated with the latest data starting late February 2024, with a full decommission expected 6 months later. Learn more about the upcoming change and where to find the new samples data in our [blog](https://waterdata.usgs.gov/blog/changes-to-sample-data/).
+Discrete water samples data are undergoing modernization, and NWIS services will no longer be updated with the latest data starting mid-March 2024, with a full decommission expected 6 months later. Learn more about the upcoming change and where to find the new samples data in our [blog](https://waterdata.usgs.gov/blog/changes-to-sample-data/).
 
 
 If you have additional questions about the qw data service, email CompTools@usgs.gov.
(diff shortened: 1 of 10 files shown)


vignettes/Status.Rmd
```diff
@@ -105,17 +105,17 @@ You'll notice that there are fewer rows in the `what_WQP` data frame (203) than
...
 
 ### Fall 2024
 
 When the NWIS services are decommissioned (possible September 2024): Should not affect WQP functions.
 
-### March/April 2024
+### April 2024
 
-When WQP gets new "data profiles" (late March/April): New data profiles will include ALL USGS data (legacy and new). Shortly after these profiles are released, we plan to update dataRetrieval so that these new profiles are the default. 
+When WQP gets new "data profiles" (April 2024): New data profiles will include ALL USGS data (legacy and new). Shortly after these profiles are released, we plan to update dataRetrieval so that these new profiles are the default.
```
lstanish-usgs, line 118: can you add something about availability of the v2 profiles? Something like: 'The WQX 2.0 profiles will continue to be available for an extended period of time.'

vignettes/Status.Rmd
```diff
@@ -21,17 +21,17 @@ This page will be updated frequently with information about the status of dataRe
 
 ### NWIS qw services
 
-Discrete water samples data are undergoing modernization, and NWIS services will no longer be updated with the latest data starting late February 2024, with a full decommission expected 6 months later. Learn more about the upcoming change and where to find the new samples data in our [blog](https://waterdata.usgs.gov/blog/changes-to-sample-data/).
```
lstanish-usgs, line 24: just to be super clear, can you update the text to read, 'and the NWIS services pulling 'qw' data will no longer...'

vignettes/Status.Rmd
lstanish-usgs: This is the link I've used for a blog post: https://exchangenetwork.net/data-exchange/wqx/

ldecicco-USGS: "The current WQP data profiles that  are considered "WQX version 2.0 profiles". These will remain available after the rollout of version 3.0. Eventually they will be retired, but there is not an estimated timeframe."

Does that sound OK? or is there a timeline?

lstanish-usgs: there is not an estimated timeframe, so this is accurate. In the past we've said something like: 'Eventually they will be retired: we will provide updated information as it becomes available.' But what you have is fine.

PR #708, opened 2024-06-10
fix legacy multipcode url

ldecicco-USGS:
And update readNWISqw documentation


lstanish-usgs:
Updates t readNWISqw documentation look good! Makes it clear that this function will be deprecated, and only delivers stale data

PR #717, opened 2024-07-24
CRAN update that only includes gwl changes

ldecicco-USGS:


R/readNWISunit.R
```diff
@@ -493,14 +514,18 @@ readNWISgwl <- function(siteNumbers,
...
     obs_url = url,
     asDateTime = TRUE,
     convertType = convertType,
     tz = tz
   )
-
+  
+  if(!is.na(parameterCd)){
+    data <- data[data$parameter_cd %in% parameterCd, ]
+  }
```
ehinman, line 527: Sorry if I missed this in our meeting: does this mean that the service sometimes pulls parameters that weren't specified in the `parameterCd` input?

lstanish-usgs:
I reviewed the diffs and everything looks good! I'm not sure how to install and test this, can that be done on this branch after the updates are merged?

ldecicco-USGS:
You could install it like this to test:

```r
remotes::install_github("ldecicco-usgs/dataRetrieval@oldCRAN")
```

PR #733, opened 2024-10-21
New defaults

ldecicco-USGS:
defaults to legacy - so shouldn't break anyone's code. But DOES offer users the chance to use the new services.


ldecicco-USGS:
@lstanish-usgs and @ehinman take a look and let me know if you have any comments.

R/constructNWISURL.R
```diff
@@ -319,7 +319,8 @@ constructNWISURL <- function(siteNumbers,
 #' retrieval for the earliest possible record.
 #' @param endDate character ending date for data retrieval in the form YYYY-MM-DD. Default is "" which indicates
 #' retrieval for the latest possible record.
-#' @param legacy Logical. If TRUE, uses legacy WQP services. Default is FALSE.
+#' @param legacy Logical. If TRUE, use legacy WQP services. Default is TRUE.
```
ehinman, line 322:
```suggestion
#' @param legacy Logical. If TRUE, uses legacy WQP services. If FALSE, uses WQX3.0 WQP services. Default is TRUE.
```

ehinman: Or maybe you don't want to offer that at this point.

R/readWQPqw.R
```diff
@@ -33,7 +33,8 @@
 #' numerics based on a standard algorithm. If false, everything is returned as a character.
 #' @param ignore_attributes logical to choose to ignore fetching site and parameter
 #' attributes. Default is \code{FALSE}.
-#' @param legacy Logical. If TRUE, use legacy WQP services. Default is FALSE.
+#' @param legacy Logical. If TRUE, use legacy WQP services. Default is TRUE.
+#' Setting legacy = FALSE uses WQX3 services, which are in-development, use with caution.
```
ehinman, line 37: Oh, I like this message here. But I see. This one is much more visible than the constructNWISURL() function.

R/readWQPqw.R
```diff
@@ -97,6 +96,11 @@ readWQPqw <- function(siteNumbers,
       retval <- create_WQP_attributes(retval, siteid = sites)
     } 
 
+    if(legacy){
+      wqp_message()
+    } else {
+      wqp_message_beta
```
ehinman, line 102: Does this one need `wqp_message_beta()`?

R/whatWQPdata.R
```diff
@@ -31,19 +32,30 @@ whatWQPsamples <- function(...,
...
   values <- sapply(values, function(x) utils::URLencode(x, reserved = TRUE))
-
-  baseURL <- drURL("ActivityWQX3", arg.list = values)
-  wqp_message_now("ActivityWQX3")
+  
+  if(legacy){
+    baseURL <- drURL("Activity", arg.list = values)
+  } else {
+    baseURL <- drURL("ActivityWQX3", arg.list = values)
+  }
```
ehinman, line 40: So in the development version, this function only performed WQX3.0 calls, but this version adds in the choice?

ehinman (commented):
These changes look good to me, just a couple curiosity questions and one slight wordsmith.

R/constructNWISURL.R
ldecicco-USGS: That text is basically in the next line. I do think changing from use to uses sounds better though.

pkgdown/templates/header.html
```diff
@@ -1,45 +1,82 @@
...
-  <![endif]-->
- 
-
-  <!-- Google Tag Manager -->
-<script>(function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
-new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
-j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
-'https://www.googletagmanager.com/gtm.js?id='+i+dl;f.parentNode.insertBefore(j,f);
-})(window,document,'script','dataLayer','GTM-TKQR8KP');</script>
-<!-- End Google Tag Manager -->
```
lstanish-usgs, line 18: Does the new page still have google tags? Having trouble locating in the changed text

ldecicco-USGS: I added that tag stuff to a new file which is injected into the header. Locally when I build the pkgdown site with the new code it's in there. Once we merge this PR, we can double-check that everything's working as expected on both GitHub and GitLab pages.

PR #735, opened 2024-10-25
Edits to status Rmd page

ehinman:
This PR updates the status page to include information about the retirement of `readNWISqw()` and the WQP functions default to WQX version 2.

### DIFF
--- a/vignettes/Status.Rmd
+++ b/vignettes/Status.Rmd
@@ -21,7 +21,7 @@ Contact CompTools@usgs.gov with additional questions.
 
 # Overview
 
-On March 11, 2024, NWIS **discrete water quality** services were "frozen": any public data retrieval using `readNWISqw()` did not include any new data. As of dataRetrieval v2.7.17, `readNWISqw()` has been removed and replaced by `readWQPqw()`. Learn more about the change and where to find the new samples data in our [blog](https://waterdata.usgs.gov/blog/changes-to-sample-data/).
+On March 11, 2024, NWIS **discrete water quality** services were "frozen": any public data retrieval using `readNWISqw()` did not include any new data. As of dataRetrieval v2.7.17, `readNWISqw()` has been retired and replaced by `readWQPqw()`. Learn more about the change and where to find the new samples data in our [blog](https://waterdata.usgs.gov/blog/changes-to-sample-data/).
 
 If you have additional questions about the NWIS qw data service, email CompTools@usgs.gov.
 
--- a/vignettes/Status.Rmd
+++ b/vignettes/Status.Rmd
@@ -73,7 +73,7 @@ knitr::kable(df)
 
 This function is generally advertised as a user-friendly function since it only works with a known list of sites, parameter codes or characterisitic names, and start/end dates. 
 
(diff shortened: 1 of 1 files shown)


vignettes/Status.Rmd
```diff
@@ -231,9 +231,9 @@ dl_data <- readWQPdata(
 ## whatNWISdata
 
 
-NWIS services are "frozen": the returned data availability will also be frozen ONLY for "qw" data_type_cd results. All other data types should not be affected. 
+NWIS discrete water quality services are "frozen": the returned data availability will also be frozen ONLY for "qw" data_type_cd results. All other data types should not be affected. 
 
-When the NWIS services are decommissioned (possible September 2024): there will no longer be any "qw" information provided in the output of `whatNWISdata`. Discrete water-quality availability will be available via WQP services. More information will be provided as we learn more.
+When the NWIS services are decommissioned (likely in 2025): there will no longer be any "qw" information provided in the output of `whatNWISdata`. Discrete water-quality availability will be available via WQP services. More information will be provided as we learn more.
```
ehinman, line 236: Actually, I think this is incorrect. I'm wondering if `whatNWISdata` still returns older qw data, or if the service HAS been decommissioned.

ldecicco-USGS: I was just looking at this. It's still showing qw data, but I'll try to get clarification on what will happen when the NWIS qw service actually gets decommissioned

lstanish-usgs: qwdata isn't slated for decommission until end of Q1, so it should still be working. Another tact for this warning is to just say that the 'qw' information has not been updated since March 2024 and will soon become deprecated: we recommend users obtain water quality data availability  from the WQP services instead. Or something

ldecicco-USGS: Right, but will there be _any_ qw info from the site inventory after Q1, or will there always be the stale info?

vignettes/Status.Rmd
```diff
@@ -73,7 +73,7 @@ knitr::kable(df)
 
 This function is generally advertised as a user-friendly function since it only works with a known list of sites, parameter codes or characterisitic names, and start/end dates.
```
lstanish-usgs, line 74: I don't see where 'this function' is defined, assuming it is readWQPqw()?

vignettes/Status.Rmd
```diff
@@ -97,11 +97,11 @@ attr(rawPcode_legacy, "url")
 
 ## readWQPdata
 
-The "readWQPdata" function is the most flexible function to get WQP data. Currently there are 4 options that use the new WQX 3.0 profiles, and 11 legacy options.
+The `readWQPdata()` function is the most flexible function to get WQP data. Currently there are 11 legacy options and 4 options that use the new WQX 3.0 profiles. Note that `readWQPdata()` does not leverage a `legacy` argument to specify which profile version the user would like returned, but instead relies on the user's specification of `service` and `dataProfile` arguments.
```
lstanish-usgs, line 100: Correction: there are 5 WQX3 profiles

vignettes/Status.Rmd
lstanish-usgs: if that info is being pulled from an rdb then yes, that info will no longer be available. IoW will not support a public endpoint for data in the qwdata DB indefinitely, at some point it will be decommissioned.

vignettes/Status.Rmd
ldecicco-USGS: It's under the header `## readWQPqw` (on line 72). We can change it though to:

The `readWQPqw` function is generally advertised...

vignettes/Status.Rmd
lstanish-usgs: Yes, the data should stop flowing when qwdata is decommissioned.

PR #738, opened 2024-11-05
Initial switch to httr2

ldecicco-USGS:


ldecicco-USGS:
@jesse-ross @ehinman @mikejohnson51 @dblodgett-usgs @mikemahoney218-usgs 
This is a pretty big PR that switches dataRetrieval from httr to httr2. I know a lot of you have been using httr2 for awhile. We're going to have some big changes to the USGS services in the next year or so, so I wanted to set things up so we could dive right in with the more modern package.

A fair number of tests were updated, but mostly it was URL attributes that the arguments got shifted around (so, same URL, different order).

If any of you have any comments, feedback, questions, let me know. I think generally everything looks to be behaving the same, except for a few cases where we had non-200 results returning without error. At the moment, those cases now automatically error. Maybe that should be the case (I think a few of our old behaviors were mostly because the service would come back with useful information even though it was a 400 or something like that). Anyway, if it's a problem, we can add a tryCatch, but I kind of think this is a more expected behavior anyway.

We're not planning an update to CRAN anytime soon. I expect this will stay in the "develop" branch for awhile.

jzemmels (approved):
These httr2 changes look good. I walked through each of the changed files and ran at least some of the examples for each function with the interactive debugger. I also ran R CMD CHECK and didn't run into issues. 

I've made some comments, but I don't think any are required to address in this PR. Approved.

DESCRIPTION
```diff
@@ -39,14 +39,14 @@ Copyright: This software is in the public domain because it contains materials
 Depends:
     R (>= 3.5.0)
```
mikemahoney218-usgs, line 40: For what it's worth, httr2 (and the tidyverse) now require R >= 4.0, and in the spring will require R >= 4.1 . If you bump this to 4.1 (which, depending when you're planning on launching things, will be the minimum version to use the package anyway) then you can use the base pipe, which might make some of the query-building more fluid

R/constructNWISURL.R
```diff
@@ -160,47 +157,45 @@ constructNWISURL <- function(siteNumbers,
...
            }
          },
          gwlevels = {
            
-           url <- appendDrURL(baseURL,
-                              site_no = siteNumbers,
-                              format = "rdb"
-           )
+           url <- httr2::req_url_query(baseURL,
+                              site_no = siteNumbers,.multi = "comma")
```
mikemahoney218-usgs, line 185:
```suggestion
                              site_no = siteNumbers, .multi = "comma")
```
Sorry, this is such a small nit but I saw it 😆

DESCRIPTION
ldecicco-USGS: 🥳I've been avoiding the 4.1 requirement, but if httr2 has it already....HELLO PIPES

R/findNLDI.R
```diff
@@ -76,21 +74,25 @@ get_nldi_sources <- function(url = pkg.env$nldi_base) {
 #' @examplesIf is_dataRetrieval_user()
 #' \donttest{
 #' base <- "https://api.water.usgs.gov/nldi/linked-data/"
-#' get_nldi(paste0(base, "comid/101"), type = "feature", use_sf = FALSE)
-#' get_nldi(paste0(base, "comid/101"), type = "feature", use_sf = TRUE)
-#' get_nldi(url = paste0(base, "nwissite/USGS-11120000"), type = "feature", use_sf = TRUE)
-#' get_nldi(paste0(base, "nwissite/USGS-11120000"), type = "feature", use_sf = TRUE)
+#' dataRetrieval:::get_nldi(paste0(base, "comid/101"), type = "feature", use_sf = FALSE)
```
mikemahoney218-usgs, line 77: I think CRAN gets mad at `:::` even if it's never actually run -- believe this gets caught by their automated checks. Maybe not an issue due to `@noRd`, but I'm not 100% sure.

If it is an issue, the workaround is to use `getFromNamespace()` to assign the non-exported object into the current environment, which defeats the static analysis checks. (If it's not an issue with CRAN, well, sorry for the distraction!)

DESCRIPTION
mikemahoney218-usgs: You could also re-export the magrittr pipe from httr2 (either directly or by adding magrittr as a direct dependency) if this is going out before the spring. Here's how httr2 does it:
https://github.com/r-lib/httr2/blob/main/R/utils-pipe.R

R/getWebServiceData.R
```diff
@@ -132,7 +115,7 @@ check_non_200s <- function(returnedList){
 default_ua <- function() {
```
mikemahoney218-usgs, line 115: have I mentioned how much I like this, by the way? This approach to tagging user agents is great

mikemahoney218-usgs:
This looks awesome 😄  Left a few minor comments (apologies everyone for the email spam -- I didn't actually intend to review the full PR tonight 😆) but this is great.

dblodgett-usgs:
🤗

R/getWebServiceData.R
```diff
@@ -1,8 +1,8 @@
...
 #'
-#' This function accepts a url parameter, and returns the raw data. The function enhances
-#' \code{\link[httr]{GET}} with more informative error messages. To add a 
-#' custom user agent, create an environmental variable: CUSTOM_DR_UA
+#' This function accepts a url parameter, and returns the raw data. 
+#' 
+#' To add a custom user agent, create an environmental variable: CUSTOM_DR_UA
 #'
 #' @param obs_url character containing the url for the retrieval
 #' @param \dots information to pass to header request
```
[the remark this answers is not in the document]
jzemmels: Sounds good. Regarding commenting, I don't think GitHub allows comments on unchanged lines ([thread](https://github.com/orgs/community/discussions/9099#discussioncomment-11336820))

mikemahoney218-usgs:
Not sure if you already know this, but just in case -- that qpdf error is an error in Homebrew, and (according to Gabor) will fix itself when the next version of the GHA runner image is deployed

R/whatWQPdata.R
```diff
@@ -31,23 +31,36 @@ whatWQPsamples <- function(...,
...
+  if(!legacy){
+    baseURL <- httr2::req_url_query(baseURL, !!!values, 
+                                    .multi = "explode")
+  } else {
+    if("siteid" %in% names(values)){
+      if(length(values[["siteid"]]) > 1){
+        sites <- values[["siteid"]]
+        sites <- paste0(sites, collapse = ";")
+        baseURL <- httr2::req_url_query(baseURL, 
+                                        siteid = sites)
```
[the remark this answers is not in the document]
ldecicco-USGS: Nice! Didn't notice we could add custom functinos.

ldecicco-USGS:
Unfortunately we're not allowed to call github actions by a version (like v2 which the r-lib folks really want us to use), so we need to occasionally include the actual commit hash in the GH action yaml file. It makes sense from a security standpoint - since the version tag is sliding - they could any anything in there and we wouldn't know.

At one point I had a script that would pull the latest commit hash from a version number, but I cannot find that anymore...if anyone on this list remembers how to do that, it would save a bit of fiddling when these things need to be updated.

PR #746, opened 2025-02-10
Add more wqx3 docs

ldecicco-USGS:
Originally taken out because the services would fail. Lots of improvements to the services since then, so we can probably start advertising these again.


ehinman (approved):
Looks good to me. I tried running all the examples. They were informative and completed successfully.

Issue #759, opened 2025-04-03
findNLDI throttling


hparker21:
**What is your question?**
I recently attempted to use the findNLDI function to find COMIDs downstream from multiple initial COMIDs. The function seemed to throttle my request in a way not previously seen. I received messages saying, "Waiting x seconds for retry backoff". It slowed down the overall process quite a bit. I have not seen this message previously when using findNLDI. Is there a way to avoid the retry backoff message and speed up the process? It now takes much longer to run a set of COMIDs through the function compared to prior attempts. 

**Expected behavior**
I have used findNLDI extensively before and it has been very fast, even when inputting lists of numerous COMIDs. I've had it process multiple entries per second in prior iterations.

dblodgett-usgs:
Apologies @hparker21 -- the NLDI has become increasingly popular and we've experienced situations where unthrottled access has caused the database to be overwhelmed and to slow down/become unresponsive. To avoid this, we've had to put rate limits in place. The client in dataRetrieval could probably be improved to avoid rate limit errors that you are experiencing, but that's about all we could do to make it better. We currently accept 400 navigation requests per hour (the limit you are hitting I think) and 3600 feature lookups per hour. 

There is a header in the request to the API that indicates if the limit is close / says that the limit has been hit... perhaps we could implement some error handling or more direct rate limiting in dataRetrieval to help out?

hparker21:
Hi David, thank you for the reply! 

Is there currently a way to view that header in the API request? If not, I think the addition of a short message notifying users if the limit has been reached or how many submissions remain before the limit is reached would be a really helpful addition. It would help us space out our requests accordingly to prevent overwhelming the API.

dblodgett-usgs:
I'll work up a PR that supports the header in dataRetrieval. Stay tuned.

dblodgett-usgs:
@hparker21 -- with what's in #762 -- you will get messages about the rate limit as you approach it and an enforced sleep will be applied if you are close to the rate limit.

[...]

PR #762, opened 2025-04-07
update throttle handling with nldi for #759

dblodgett-usgs:
I'm not totally satisfied with this, but I think it ticks the box for #759

I now see: 

```r
> sapply(1:500, function(x) findNLDI(comid = 101, nav = "DM", find = c("nwis", "wqp")))
Approaching NLDI rate limit. 28 of 400 remaining
Approaching NLDI rate limit. 26 of 400 remaining
Approaching NLDI rate limit. 25 of 400 remaining
Approaching NLDI rate limit. 23 of 400 remaining
Approaching NLDI rate limit. 22 of 400 remaining
Approaching NLDI rate limit. 20 of 400 remaining
Approaching NLDI rate limit. 19 of 400 remaining
Sleeping to try to avoid going over rate limit.
Approaching NLDI rate limit. 19 of 400 remaining
Sleeping to try to avoid going over rate limit.
Approaching NLDI rate limit. 20 of 400 remaining
Approaching NLDI rate limit. 19 of 400 remaining
Sleeping to try to avoid going over rate limit.
```

This won't send messages unless a user gets to within 90% of the overall rate limit. It won't sleep until the user is under 20 requests remaining in their allocation. The rate limit calculations are a black box so I was quite conservative with waiting 3x what should be technically possible. It seems like the rate limits accrue in some way differently from 1:1. 

The way the rate limit header is listed, it shows the absolute limit and how many is left in that allocation. It does not say anything about the time base for the rate limit. For now, it's all based on a per hour allocation. We will try to stick to the hour basis going forward.

### DIFF
--- a/R/findNLDI.R
+++ b/R/findNLDI.R
@@ -45,9 +45,7 @@ find_good_names <- function(input, type) {
 get_nldi_sources <- function(url = pkg.env$nldi_base) {
   res <- httr2::request(url)
   res <- httr2::req_user_agent(res, default_ua())
-  res <- httr2::req_throttle(res, rate = 30 / 60) 
-  res <- httr2::req_retry(res, 
-                          backoff = ~ 5, max_tries = 3) 
+  res <- httr2::req_error(res, is_error = \(x) FALSE)
   res <- httr2::req_perform(res)
 
   if (res$status_code == 200) {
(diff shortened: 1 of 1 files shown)


dblodgett-usgs:
@ldecicco-USGS -- this is my first time touching the httr2 implementation in here -- hope I didn't step in anything.

ldecicco-USGS:
Looks good!

PR #773, opened 2025-05-16
New API services!

ldecicco-USGS:
First pass at:

read_NWIS_daily
read_NWIS_monitoring_location
read_NWIS_ts_meta

To use the inline R functionality in the roxygen2 (to take advantage of dynamically pulling in the param descriptions), I had to switch the roxygen from classic to markdown...which is why there are 75 changed files.


ldecicco-USGS:
I'll work on getting a API key in the repo next week.

Other things that are known to be lacking, on the TODO list, but first hoping to get some functional stuff up for users to test:
1. No tests. All the functions have working examples, which is kind of a test, but once we're happy with the flow, we can add some tests on the new functions.
2. Internal functions don't have documentation. 
3. Vignette needs work!

jzemmels (changes requested):
Two edits to the pkgdown.yml file to make sure the website loads. I'll add more commentary about the actual function changes soon. In the meantime, I wanted to ask:

- Do we want to update the tests, internal docs, and vignettes in this PR, or were you planning on doing that in a follow-up? If the former, I'm happy to start chipping away.
- What do we think about adding functions like `read_USGS_dv <- read_USGS_daily` to bridge the old vs. new functionality a bit more?

ldecicco-USGS:
1. I was _planning_ to get this PR reviewed while simultaneously working on a new PR with those things, but whatever is most efficient is fine. That being said, let's not spend too much time chipping away at the non-user-side issues - there's going to be a blog out introducing the new services soon - it would be great to have the "develop" branch at a state that users can give it a try. I will request and push up a CI API token though to get the pipelines to pass without hitting the limits.

2. I'd vote strongly to not create or include a function that helps folks read_USGS_dv <- read_USGS_daily

We can/will show examples on how to do it in the vignette. There are several reasons not to make a function... mostly because it's another function that we would need to maintain over time that encourages using wide data frames instead of long data frames. We'd need to do the conversion in base R if we made the function (or add tidyr to the imports which we've tried VERY hard not to do), which isn't that bad, but still trickier and looks complicated (if users look at the function for inspiration for how they should do their conversions). If we keep the directions in the vignettes, we can just use the pivot_wider function which I'm confident we can present in a clear/approachable way so that folks understand how to use it. We should also focus on providing examples to show why this new format is far superior to the legacy one.

jzemmels:
> 1. I was _planning_ to get this PR reviewed while simultaneously working on a new PR with those things, but whatever is most efficient is fine. That being said, let's not spend too much time chipping away at the non-user-side issues - there's going to be a blog out introducing the new services soon - it would be great to have the "develop" branch at a state that users can give it a try. I will request and push up a CI API token though to get the pipelines to pass without hitting the limits.
> 2. I'd vote strongly to not create or include a function that helps folks read_USGS_dv <- read_USGS_daily
> 
> We can/will show examples on how to do it in the vignette. There are several reasons not to make a function... mostly because it's another function that we would need to maintain over time that encourages using wide data frames instead of long data frames. We'd need to do the conversion in base R if we made the function (or add tidyr to the imports which we've tried VERY hard not to do), which isn't that bad, but still trickier and looks complicated (if users look at the function for inspiration for how they should do their conversions). If we keep the directions in the vignettes, we can just use the pivot_wider function which I'm confident we can present in a clear/approachable way so that folks understand how to use it. We should also focus on providing examples to show why this new format is far superior to the legacy one.

Sounds good -- we can work from the other PR you're planning if you'd like to get these changes merged quickly.

For point 2, I think I was non-specific about what I was proposing. What I asking about is having a function called `read_USGS_dv` that is a carbon copy of `read_USGS_daily`, just under a different name. As described here: https://stackoverflow.com/a/57771080.

vignettes/read_USGS_functions.Rmd
```diff
@@ -0,0 +1,185 @@
...
+```{r}
+library(ggplot2)
+ggplot(data = dv_modern) +
+  geom_point(aes(x = time, y = value, 
+                 color = approval_status)) +
+  facet_grid(parameter_code ~ ., scale = "free") +
+  theme_bw()
+
+```
+
```
jzemmels, line 77: (SC) It might be helpful to provide users with a bit of code for transforming the new long format back to something like the wide format (unless we want to force them to adopt the new format). I think the following code gets most of the way there. There may be qualifier codes not yet represented.

``` r
dv_modern |> 
  sf::st_drop_geometry() |>
  dplyr::mutate(qualifier = dplyr::case_when(approval_status == "Provisional" ~ "P",
                                             approval_status == "Approved" & qualifier == "Estimated" ~ "A e",
                                             approval_status == "Approved" ~ "A",
                                             .default = NA)) |>
  tidyr::pivot_wider(names_from = c(parameter_code, 
                                    statistic_id), 
                     names_glue = "X_{parameter_code}_{statistic_id}{ifelse(.value == 'qualifier', '_cd', '')}", 
                     values_from = c(value, qualifier),
                     id_cols = c(time, monitoring_location_id),
                     names_vary = "slowest") |>
  dplyr::arrange(time, monitoring_location_id) |>
  tidyr::separate_wider_delim(cols = monitoring_location_id, 
                              delim = "-",
                              names = c("agency_cd", "site_no")) |>
  dplyr::rename(Date = time) |>
  dplyr::select(agency_cd, site_no, Date, tidyselect::starts_with("X_")) 
```

vignettes/read_USGS_functions.Rmd
```diff
@@ -0,0 +1,185 @@
...
+As we bid adieu to the NWIS web services, we welcome a new web service offering, the USGS Water Data OGC APIs.
+
+# New USGS data access
+
+This is a modern access point for USGS data. The USGS is planning to modernize all web services in the near future. For each of these updates, `dataRetrieval` will create a new function to access the new services. 
+
+# Daily Values
+
+To access these services on a web browser, go to <https://api.waterdata.usgs.gov/ogcapi/v0/collections/daily>.
```
jzemmels, line 43: (SC) I know this PR isn't intended to have complete vignettes, but I'll add some thoughts for the future: it'd be great to provide users with a description of each service at the beginning. "This service provides [x]"

vignettes/read_USGS_functions.Rmd
```diff
@@ -0,0 +1,185 @@
...
+You can use `usethis::edit_r_environ()` to edit find and open your .Renviron file. You will need to restart R for that variable to be recognized. 
+
+
+# Notes on dataRetrieval development
+
+## New Features
+
+### Style
+
+New functions will use a "snake case", such as "read_USGS_samples". Older functions use camel case, such as "readNWISdv". The difference is the underscore between words. This should be a handy way to tell the difference between newer modern data access, and the older traditional functions.
```
jzemmels, line 176: (SC) I personally prefer snake_case, but I'm wondering if we could create carbon copies of these functions that use camelCase: `readUSGSsamples`? Or maybe that doesn't really matter -- just a thought.

jzemmels (changes requested):
Some quirky behavior I noticed while running the examples on VPN without an API key and documentation recommendations. I've separated my comments into Required Changes (RC, meaning something I think should be addressed before merging), of which there's only one, and Suggested Changes (SC).

Another question: do we want to print a message .onLoad if the user does not have an API key set?

jzemmels:
I just noticed that running these functions with 0 arguments will start running successfully. I'm not sure if it's pulling literally all available data in the endpoint? I didn't wait to find out, but perhaps we should add a check at the beginning that the user has specified *something*.

ldecicco-USGS:
Ha, good call. I can get a check up early this afternoon if you don't push up any suggestions. 

As for how many name iterations of functions to add, I think ultimately people will get use to the new names. I like keeping the new ones snake and the old one camel because it's a visual way to describe if the function is new or old. Since I advocate in intro to R classes a lot to use the "tab" key for autocomplete, I wonder if making alternative names will ultimately be more or less confusing? But **if** we do decide to do it, I'd vote just the dv -> daily, site -> monitoring_location, and as the new services come up the old suffix -> new suffix

R/walk_pages.R
```diff
@@ -0,0 +1,100 @@
...
+  if(!is.null(req$body)){
+    url_method <- "POST"
+    body <- req$body
+  }
+  
+  use_sf <- !grepl("skipGeometry=true", req$url, ignore.case = TRUE)
+  
+  message(url_method, ": ", req$url) 
+  
+  returned_contents <- httr2::req_perform(req)
```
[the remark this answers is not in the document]
ldecicco-USGS: Good call, the `httr2::req_error` will give us a lot of nice options.

vignettes/read_USGS_functions.Rmd
jzemmels: Ahh, I missed that long_to_wide article. I think that provides a good overview that would be redundant to mention here. I agree that this vignette should focus on the new bells and whistles that come with the new functions.

R/construct_api_requests.R
```diff
@@ -0,0 +1,296 @@
...
+#' 
+construct_api_requests <- function(service,
+                                   properties = NA_character_,
+                                   bbox = NA,
+                                   crs = NA_character_,
+                                   limit = 10000,
+                                   skipGeometry = FALSE,
+                                   ...){
+  
+  schema <- check_OGC_requests(endpoint = service,
```
jzemmels, line 53:
```suggestion
  if(!any(unlist(lapply(list(properties, bbox, crs, skipGeometry, ...), isTRUE)))){
    choice <- menu(choices = c("Yes", "No"),
                   title = paste0("You have not set any arguments, meaning you're about to request *all* records available in the ", service," service. Are you sure you want to proceed?"))
    if(choice == 2){
      stop("Stopping function execution.")
    }
  }

  schema <- check_OGC_requests(endpoint = service,
```

ldecicco-USGS: @mikemahoney218-usgs I personally can't imagine that if a user would request all-the-data in good faith. Do you think it's OK to give them an **error** if they haven't requested any filters, or is that something your group would expect/support? Since there are built-in limits, it shouldn't timeout...but there will be bajillions of pages that would come back.

jzemmels: I don't mind adding an `interactive()` to the if conditional, though a first-time user just trying to acquaint themselves with  the service might bristle at being given a prompt the first time they run the function. Maybe simply printing a warning would have a similar effect while being less obtrusive?

mikemahoney218-usgs: Oh, I also just realized -- there will be some services (/parameter-codes comes to mind as the first example) where maybe people _do_ actually want to loop through the full data set. None of the ones that exist yet, but once we're putting reference lists through the APIs that might be a real use case.

jzemmels: Well, maybe we table this topic for later. We can monitor the logs to determine how often people are requesting all the data (as part of a larger dataRetrieval user study!!!!) and assess whether further action is needed.

ldecicco-USGS: Or something like:
```r
check_n_arguments <- function(...){
  if(!any(unlist(lapply(list(...), isTRUE)))){
    if(interactive()){
      choice <- menu(choices = c("Yes", "No"),
                     title = paste0("You have not set any arguments, meaning you're about to request *all* records available in the ", service," service. Are you sure you want to proceed?"))
      if(choice == 2){
        stop("Stopping function execution.")
      }    
    } else {
      message("No filtering arguments set.")
    }
  }
}
```
And we don't add that to the eventual pcode service? 

OR, just make it a warning/message? If a user is in interactive mode, presumably they can stop it after they notice the warning. 

```r
if(!any(unlist(lapply(list(properties, bbox, crs, skipGeometry, ...), isTRUE)))){
  warning("No arguments, have fun waiting!")
}
```

jzemmels: I'm currently leaning towards the warning message (especially the cheeky one currently proposed :) ). But I can be swayed.

vignettes/read_USGS_functions.Rmd
ldecicco-USGS: I added the descriptions pulled directly from the services:

```
# Daily Values

`r dataRetrieval:::get_description("daily")`

To access these services on a web browser, go to
```

vignettes/read_USGS_functions.Rmd
ldecicco-USGS: Tabling this for now, hoping most users are learning to tab.

R/construct_api_requests.R
```diff
@@ -0,0 +1,296 @@
...
+  
+  template_path_post <- system.file("templates/post.CQL2", package = "dataRetrieval")
+  template_post <- readChar(template_path_post, file.info(template_path_post)$size)
+  
+  single_params <- c("datetime", "last_modified", "begin", "end", "time")
+  
+  full_list <- list(...)
+  get_list <- full_list[names(full_list) %in% single_params]
+  
+  get_list[["crs"]] <- crs
```
mikemahoney218-usgs, line 90: Just want to note that I'm working on building out better CRS support... you specify this using a URL to one of the CRS supported by a given collection. So for instance, on the daily collection page:

https://api.waterdata.usgs.gov/ogcapi/v0/collections/daily?f=json

You'll notice a JSON blob:
```json
    "crs":[
        "http://www.opengis.net/def/crs/OGC/1.3/CRS84",
        "http://www.opengis.net/def/crs/OGC/1.3/CRS83",
        "http://www.opengis.net/def/crs/OGC/1.3/CRS27"
    ],
```

Those correspond to WGA84, NAD83, and NAD27. We can't use the EPSG names for these yet because... I'm just gonna link https://github.com/geopython/pygeoapi/issues/2010 

_so_ I can try and put together better docs for this somewhere, but wanted to flag that the CRS argument doesn't work like anything users would expect... yet, at any rate

R/construct_api_requests.R
```diff
@@ -0,0 +1,296 @@
...
+  
+  post_list <- full_list[!names(full_list) %in% single_params]
+  
+  post_params <- explode_post(post_list)
+  
+  if(length(post_params) > 0){
+    POST = TRUE
+  }
+  
+  time_periods <- c("last_modified", "datetime", "time")
```
mikemahoney218-usgs, line 102: also "begin" and "end" on time-series-metadata

R/construct_api_requests.R
```diff
@@ -0,0 +1,296 @@
...
+  baseURL <- httr2::request("https://api.waterdata.usgs.gov/ogcapi/v0/collections") |> 
+    httr2::req_url_path_append(service, "items") |> 
+    basic_request() 
+  
+}
+
+format_api_dates <- function(datetime){
+  
+  if(!any(isTRUE(is.na(datetime)) | isTRUE(is.null(datetime)))){
+    if(length(datetime) == 1){
```
mikemahoney218-usgs, line 153: will this work for duration requests -- eg, `time=P1D`?

R/construct_api_requests.R
```diff
@@ -0,0 +1,296 @@
...
+    } else {
+      stop("datetime should only include 1-2 values")
+    }
+  }
+  return(datetime)
+}
+
+explode_post <- function(ls){
+  
+  ls <- Filter(Negate(anyNA), ls)
```
mikemahoney218-usgs, line 171: love seeing Negate() in the wild

R/construct_api_requests.R
```diff
@@ -0,0 +1,296 @@
...
+  parameters <- paste0(unlist(parameter), collapse = '", "')
+  parameters <- paste0('"', parameters, '"')
+  parameter_list <- list("property" = names(parameter),
+                         "parameter" = parameters)
+  return(whisker::whisker.render(template, parameter_list))
+} 
+
+#' Check OGC requests
+#' 
+#' @param endpoint Character, can be "daily", "timeseries-metadata"
```
mikemahoney218-usgs, line 200:
```suggestion
#' @param endpoint Character, can be "daily", "time-series-metadata", "monitoring-locations"
```

Might be worth genericizing to "can be any existing collection"?

R/read_USGS_daily.R
```diff
@@ -0,0 +1,112 @@
...
+                         crs = NA_character_,
+                         skipGeometry = NA,
+                         time = NA_character_,
+                         convertType = TRUE){
+  
+  message("Function in development, use at your own risk.")
+  
+  service <- "daily"
+  
+  dv_req <- construct_api_requests(service = service,
```
mikemahoney218-usgs, line 82: for what it's worth, I've used a much less readable but self-maintaining approach with do.call in the past
https://github.com/Permian-Global-Research/rsi/blob/main/R/get_stac_data.R#L702-L705

R/walk_pages.R
```diff
@@ -0,0 +1,102 @@
...
+    }
+    
+    return(return_df)
+    
+  } else {
+    
+    make_request <- httr2::req_url(req = req, 
+                                   url = json_content$links[, "href", drop = TRUE][next_page])
+    
+    Tailcall(
```
mikemahoney218-usgs, line 95: I just saw that Tailcall is an R 4.4 feature -- might be worth moving to a loop instead to keep backwards compatibility?

R/construct_api_requests.R
ldecicco-USGS: Any user that knows what crs is should know how to convert the geometry once they have it, right? I'm tempted to take that argument out (at least for now?)

mikemahoney218-usgs (commented):
realized too late I was slowly working through the full code and should actually group comments into a review instead

R/construct_api_requests.R
mikemahoney218-usgs: I'd say that anyone who knows that there are CRS listed at opengis.net is, definitionally, a power user who can handle converting for themselves after the fact

R/read_USGS_daily.R
ldecicco-USGS: Slick

R/walk_pages.R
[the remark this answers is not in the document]
ldecicco-USGS: I added:
```
dataRetrieval 2.7.18.9002
Extended Documentation: https://doi-usgs.github.io/dataRetrieval 
Consider adding an API_USGS_PAT for new USGS functions.
See: https://api.waterdata.usgs.gov/signup
```
(and those last 2 lines shouldn't show up if the API_USGS_PAT is set.

I have the startup messages on `.onAttach`. I _think_ that's better because it only shows on `library(dataRetrieval)` and not `dataRetrieval::read_USGS_daily`, but maybe I missed a memo...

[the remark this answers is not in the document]
jzemmels: Looks great! In this case, I'm fine with `.onAttach`.

R/walk_pages.R
ldecicco-USGS: "Tailcall and Exec are experimental and may be changed or dropped in future released versions of R."

R/read_USGS_monitoring_location.R
```diff
@@ -0,0 +1,137 @@
...
+#'                                             "state_name",
+#'                                             "country_name"))
+#'
+#' site_info_no_sf <- read_USGS_monitoring_location(monitoring_location_id = site,
+#'                                    skipGeometry = TRUE)
+#' 
+#' multi_site <- read_USGS_monitoring_location(state_name = "Wisconsin")
+#' 
+#' }
+read_USGS_monitoring_location <- function(monitoring_location_id = NA_character_,
```
ehinman, line 77: Ran into an issue with this call (is it the query rate limiter or...?):

```
sites = whatNWISsites(statecd = "NY")
sites = paste0("USGS-", sites$site_no)

blah <- read_USGS_monitoring_location(monitoring_location_id = sites, skipGeometry = TRUE)
Function in development, use at your own risk.
POST: https://api.waterdata.usgs.gov/ogcapi/v0/collections/monitoring-locations/items?f=json&lang=en-US&skipGeometry=TRUE&limit=10000
Error in `httr2::req_perform()` at dataRetrieval/R/walk_pages.R:53:3:
! HTTP 403 Forbidden.
Run `rlang::last_trace()` to see where the error occurred.
> rlang::last_trace()
<error/httr2_http_403>
Error in `httr2::req_perform()` at dataRetrieval/R/walk_pages.R:53:3:
! HTTP 403 Forbidden.
---
Backtrace:
    ▆
 1. └─dataRetrieval::read_USGS_monitoring_location(...)
 2.   └─dataRetrieval:::walk_pages(site_req) at dataRetrieval/R/read_USGS_monitoring_location.R:132:3
 3.     └─dataRetrieval:::walk_pages_recursive(req = req, page = 1, contents = list()) at dataRetrieval/R/walk_pages.R:34:3
 4.       └─httr2::req_perform(req) at dataRetrieval/R/walk_pages.R:53:3
Run rlang::last_trace(drop = FALSE) to see 3 hidden frames.
```
Just interesting that if I use `sites[1:100]` as an input, it works fine, even with a POST. `sites` is 88,922 sites long, so maybe I'm hitting the limit of what the database accepts.

R/read_USGS_monitoring_location.R
```diff
@@ -0,0 +1,137 @@
...
+                            agency_code = NA_character_,
+                            agency_name = NA_character_,
+                            monitoring_location_number = NA_character_,
```
ehinman, line 80: Interesting that there's `id` and `number`. My guess is that hardcore legacy users will want to stick with `number`.

R/read_USGS_daily.R
```diff
@@ -0,0 +1,102 @@
...
+#'                         parameter_code = "00060",
+#'                         time = "P7D")
+#' 
+#' multi_site <- read_USGS_daily(monitoring_location_id =  c("USGS-01491000", 
+#'                                                        "USGS-01645000"),
+#'                         parameter_code = c("00060", "00010"),
+#'                         time = c("2023-01-01", "2024-01-01"))
+#' 
+#' }
+read_USGS_daily <- function(monitoring_location_id = NA_character_,
```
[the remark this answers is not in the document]
ehinman: Hmmm....also here. I first just put in a YYYY-MM-DD and got an error, consulted the help, then tried a single datetime, and that failed, then tried a bounded interval:

```
site_test <- read_USGS_daily(monitoring_location_id = "USGS-06137570", parameter_code = '00060', last_modified = "2018-02-12T00:00:00Z/2024-03-18T12:31:12Z")
Function in development, use at your own risk.
GET: https://api.waterdata.usgs.gov/ogcapi/v0/collections/daily/items?f=json&lang=en-US&limit=10000&monitoring_location_id=USGS-06137570&parameter_code=00060&last_modified=2018-02-12T00%3A00%3A00Z%2F2024-03-18T12%3A31%3A12Z
Error in order(return_list$time, return_list$monitoring_location_id) : 
  argument 1 is not a vector
In addition: Warning messages:
1: In min(bb[, 1L], na.rm = TRUE) :
  no non-missing arguments to min; returning Inf
2: In min(bb[, 2L], na.rm = TRUE) :
  no non-missing arguments to min; returning Inf
3: In max(bb[, 3L], na.rm = TRUE) :
  no non-missing arguments to max; returning -Inf
4: In max(bb[, 4L], na.rm = TRUE) :
  no non-missing arguments to max; returning -Inf
5: Unknown or uninitialised column: `time`. 
6: Unknown or uninitialised column: `monitoring_location_id`. 
```

ehinman (commented):
Still doing some testing and such, but here are a few thoughts/discoveries.

R/read_USGS_daily.R
```diff
@@ -0,0 +1,102 @@
...
+#' @param unit_of_measure `r get_params("daily")$unit_of_measure`
+#' @param approval_status `r get_params("daily")$approval_status`
+#' @param last_modified `r get_params("daily")$last_modified`
+#' @param time_series_id `r get_params("daily")$time_series_id`
+#' @param qualifier `r get_params("daily")$qualifier`
+#' @param daily_id `r get_params("daily")$id`
+#' @param properties The properties that should be included for each feature.
+#' The parameter value is a comma-separated list of property names. Available options are
+#' `r schema <- check_OGC_requests(endpoint = "daily", type = "schema"); paste(names(schema$properties), collapse = ", ")`
+#' @param bbox Only features that have a geometry that intersects the bounding
```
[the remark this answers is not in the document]
mikemahoney218-usgs: xmin,ymin,xmax,ymax

I know this is specified somewhere... but I don't have the standard number to hand.

Something that might be wroth flagging is that sf::st_bbox will automatically be in the right sequence (because it uses the same standard)

R/construct_api_requests.R
```diff
@@ -0,0 +1,301 @@
...
+#' 
+#' sites <- c("USGS-01491000", "USGS-01645000")
+#' start_date <- "2018-01-01"
+#' end_date <- "2022-01-01"
+#' req_dv <- construct_api_requests("daily",
+#'                                 monitoring_location_id = sites,
+#'                                 parameter_code = c("00060", "00065"),
+#'                                 datetime = c(start_date, end_date))
+#' 
+construct_api_requests <- function(service,
```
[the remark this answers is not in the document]
mikemahoney218-usgs: `monitoring_location_id` is the full parameter -- you partially matched it with the second example

We use ID instead of identifier throughout because of the geojson standard suggesting "id" as a top-level parameter. It's a bit of an infelicity though, considering we use very few abbreviations otherwise

R/read_USGS_monitoring_location.R
mikemahoney218-usgs: Can I ask why you're querying for a specific 88,922 sites? I have a hard time imagining the use case for that one 😅

R/read_USGS_monitoring_location.R
mikemahoney218-usgs: Data furnished from other agencies can have the same site number as USGS sites. Monitoring location IDs are guaranteed to be unique across the entire network.

R/construct_api_requests.R
```diff
@@ -0,0 +1,301 @@
...
+#' Swagger docs: <https://api.waterdata.usgs.gov/ogcapi/v0/openapi?f=html>.
+#' 
+#' @export
+#' @param service Which service
+#' @param ... Extra parameters from the specific services.
+#' @param bbox Only features that have a geometry that intersects the bounding
+#' box are selected.The bounding box is provided as four or six numbers, depending
+#' on whether the coordinate reference system includes a vertical axis (height or
+#' depth).
+#' @param properties The properties that should be included for each feature. The
```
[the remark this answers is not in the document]
mikemahoney218-usgs: Yes, exactly that! GeoJSON stashes attributes under the "properties" key, so if you're making direct API calls it makes sense that you'd filter that array down using the properties argument.

R/walk_pages.R
mikemahoney218-usgs: yeah, I'm personally betting that note will be there in 20 years 😆

R/walk_pages.R
```diff
@@ -0,0 +1,102 @@
...
+  
+  if("value" %in% names(df)){
+    df$value <- as.numeric(df$value)
+  }
+  df
+}
+
+walk_pages <- function(req) {
+  
+  walk_pages_recursive(
```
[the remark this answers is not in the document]
mikemahoney218-usgs: Pretty sure this is derived from the version I sent Laura a while back -- and I _think_ I set it up this way so that you can initialize `contents` and `page` outside of the recursive context, so they don't get reset and you don't need global assignment. I _think_

R/read_USGS_monitoring_location.R
ehinman: Makes sense.

R/read_USGS_monitoring_location.R
ehinman: LOL. I didn't realize you could query by `state_name`. Shame on me. Probably unlikely.

jzemmels: But another good reason for `req_error()` handling!

R/construct_api_requests.R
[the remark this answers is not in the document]
ldecicco-USGS: Whoops sorry about that, pushed up a change yesterday and didn't test the bbox for side affects

R/construct_api_requests.R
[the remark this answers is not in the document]
ldecicco-USGS: I toyed initially with the idea of calling the argument "columns", but don't want to sway too far from the actual service. How about we be more explicit in the descriptions, but still call the actual argument "properties"?

```
#' @param properties A vector of requested columns to be returned from the query.
#' Available options are: 
```

R/read_USGS_daily.R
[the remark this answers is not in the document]
ldecicco-USGS: For the first example, you asked for a stat code that the site doesn't offer - so a great example of a valid request with no returned data. I'll work up a better response. It's something to think about.

1. The service itself gives some json:

```
{
    "type":"FeatureCollection",
    "features":[],
    "numberReturned":0,
    "links":[
        {
            "type":"application/geo+json",
            "rel":"self",
            "title":"This document as GeoJSON",
            "href":"https://api.waterdata.usgs.gov/ogcapi/v0/collections/daily/items?f=json&lang=en-US&limit=10000&monitoring_location_id=USGS-06137570&parameter_code=00060&statistic_id=00002"
        },
        {
            "rel":"alternate",
            "type":"application/ld+json",
            "title":"This document as RDF (JSON-LD)",
            "href":"https://api.waterdata.usgs.gov/ogcapi/v0/collections/daily/items?f=jsonld&lang=en-US&limit=10000&monitoring_location_id=USGS-06137570&parameter_code=00060&statistic_id=00002"
        },
        {
            "type":"text/html",
            "rel":"alternate",
            "title":"This document as HTML",
            "href":"https://api.waterdata.usgs.gov/ogcapi/v0/collections/daily/items?f=html&lang=en-US&limit=10000&monitoring_location_id=USGS-06137570&parameter_code=00060&statistic_id=00002"
        },
        {
            "type":"application/json",
            "title":"Daily values",
            "rel":"collection",
            "href":"https://api.waterdata.usgs.gov/ogcapi/v0/collections/daily"
        }
    ],
    "timeStamp":"2025-05-23T14:18:51.680649Z"
}
```

Which, we can pretty easily check for the `"numberReturned":0,`. HOWEVER, historically dataRetrieval users really like their empty requests to come back with an empty data frame so they can have it in a big ol' loop and just keep binding rows together. 

Side note: I'm about to convert all the readNWISdv and readNWISsite functions in the testthat tests to these functions - so there will probably be other error handling stuff like this fixed today.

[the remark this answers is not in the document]
ldecicco-USGS: Similarly that 2nd example had no data that was last modified between 2018 and 2024. If you changed it to:

```r
site_test <- read_USGS_daily(monitoring_location_id = "USGS-06137570", 
                             parameter_code = '00060',
                             last_modified = "2024-03-18T12:31:12Z/2025-05-18T12:31:12Z")
```
then you'd get some results. So, number 1 priority this morning is the no data result. Give me a little bit to think about that....

ehinman:
> AH! Yeah, that sounds cool!

I'm not sure if I think this is more confusing to have these "shadow" functions. The pros are that it keeps some "normalcy" for what was, but the cons are that it keeps people from fully adopting the new function. Feel free to push back, I just would rather limit the number of available options.

ldecicco-USGS:
> > AH! Yeah, that sounds cool!
> 
> I'm not sure if I think this is more confusing to have these "shadow" functions. The pros are that it keeps some "normalcy" for what was, but the cons are that it keeps people from fully adopting the new function. Feel free to push back, I just would rather limit the number of available options.

Yup, I think after some back-and-forth scattered above I've talked myself out of it too.

jzemmels:
> > AH! Yeah, that sounds cool!
> 
> I'm not sure if I think this is more confusing to have these "shadow" functions. The pros are that it keeps some "normalcy" for what was, but the cons are that it keeps people from fully adopting the new function. Feel free to push back, I just would rather limit the number of available options.

I consider it as being the new function, just with a name that evokes the old function that may be easier for users to remember. I'll admit that my desire to add this is based on a frustration I sometimes have with `tidyverse` and other packages where they change a function's based on what *they* consider best. It took me a long time to remember `pivot_wider` and `pivot_longer` when I had originally learned `gather` and `spread` -- and I'm sure those functions frustrated the `melt` and `cast` crowd. 

I'm thinking that long-time `dataRetrieval` users will first remember `dv` when they're trying to find something before they remember that it's now called `daily`. Plus it's "cheap" to add a function like this since it's literally just a carbon copy of another.

R/read_USGS_monitoring_location.R
mikemahoney218-usgs: Agree -- but our suggestion to this user is going to be "consider splitting your request into multiple pieces"

R/read_USGS_daily.R
```diff
@@ -0,0 +1,105 @@
...
+#' @param qualifier `r get_params("daily")$qualifier`
+#' @param daily_id `r get_params("daily")$id`
+#' @param properties A vector of requested columns to be returned from the query.
+#' Available options are: 
+#' `r schema <- check_OGC_requests(endpoint = "daily", type = "schema"); paste(names(schema$properties), collapse = ", ")`
+#' @param bbox Only features that have a geometry that intersects the bounding
+#' box are selected.The bounding box is provided as four or six numbers, depending
+#' on whether the coordinate reference system includes a vertical axis (height or
+#' depth). Coordinates are assumed to be in crs 4326.
+#' @param limit The optional limit parameter limits the number of items that are
```
ehinman, line 24: I'm confused about what this means. I tried this:

```
bbox <- read_USGS_daily(bbox = c(-83, 36.5, -81, 38.5), parameter_code = c("00060", "00065"), limit = 100)
```
And it was super fast, and returned exactly 2000 rows, which seemed weird, but I'll roll with it. 

I then tried this (no limit):

```
bbox <- read_USGS_daily(bbox = c(-83, 36.5, -81, 38.5), parameter_code = c("00060", "00065"))
```
And it cannot seem to complete. I tried running it twice. Here's the error I received:
Screenshot 2025-05-27 at 11 42 51 AM
Screenshot 2025-05-27 at 11 41 52 AM

R/read_USGS_daily.R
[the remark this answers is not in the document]
ehinman: These all work great now, with the first one returning a df of 0 rows. Thanks @ldecicco-USGS!!

R/read_USGS_daily.R
```diff
@@ -0,0 +1,105 @@
...
+  
+  return_list <- deal_with_empty(return_list, properties, service)
+    
+  if(convertType) return_list <- cleanup_cols(return_list)
+
+  return_list <- return_list[order(return_list$time, return_list$monitoring_location_id), ]
+  
+  return_list <- rejigger_cols(return_list, properties, service)
+  
+  return(return_list)
```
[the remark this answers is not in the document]
ehinman: Ahhh. Hmm....are pages organized in a consistent way? (a question for @mikemahoney218-usgs) Like say someone makes a big call that eventually times out after X pages have been completed...is there any way they can use the information from the completed pages to simply "continue" on downloading in a new call?

R/construct_api_requests.R
```diff
@@ -0,0 +1,318 @@
...
+  
+}
+
+error_body <- function(resp) {
+  status <- httr2::resp_status(resp)
+  if(status == 429){
+    x <- httr2::resp_body_json(resp)$error
+    return(x[["message"]])
+  } else if (status == 403){
+    return("Query request denied. Possible reasons include query being exceeding server limits.")
```
ehinman, line 263:
```suggestion
    return("Query request denied. Possible reasons include query exceeding server limits.")
```

R/read_USGS_monitoring_location.R
ehinman: Message looks good upon retrying it. Thanks!

R/construct_api_requests.R
```diff
@@ -0,0 +1,318 @@
...
+                                   limit = 10000,
+                                   skipGeometry = FALSE,
+                                   ...){
+  
+  schema <- check_OGC_requests(endpoint = service,
+                               type = "schema")
+  all_properties <- names(schema$properties)
+  
+  match.arg(properties, choices = c(all_properties, NA_character_),
+            several.ok = TRUE)
```
ehinman, line 55: This is minor, so feel free to disregard unless it becomes a problem. But, nonetheless something I noticed while crashing around in here.

```
site_test <- read_USGS_daily(monitoring_location_id = "USGS-06137570", parameter_code = '00060', properties = c("dog"))
```
I love the message that tells you what you can place in `properties`. However, one of the options is displayed as "NA". So I decided to try "NA", like this:

```
site_test <- read_USGS_daily(monitoring_location_id = "USGS-06137570", parameter_code = '00060', properties = c("NA"))
```
...and it gave me a separate error:
```
Error in match.arg(properties, choices = c(all_properties, NA_character_),  : 
  'arg' must be NULL or a character vector
```
At this point, someone should know to replace "NA" with NA_character_....and really just shouldn't mess with it at all if they want everything. I guess my point is that maybe "NA" shouldn't be in the list of allowable values, since it needs to be that specific NA_character_.

R/read_USGS_daily.R
ehinman: I guess I'm still confused on what `limit` means, though. My limit was 100 and I got 2000 rows returned. What is the math on that?

R/read_USGS_daily.R
```diff
@@ -0,0 +1,105 @@
...
+#'                         
+#' dv_data_period <- read_USGS_daily(monitoring_location_id = site,
+#'                         parameter_code = "00060",
+#'                         time = "P7D")
+#' 
+#' multi_site <- read_USGS_daily(monitoring_location_id =  c("USGS-01491000",
+#'                                                        "USGS-01645000"),
+#'                         parameter_code = c("00060", "00010"),
+#'                         limit = 500,
+#'                         time = c("2023-01-01", "2024-01-01"))
```
ehinman, line 61: Ah, so here's an example (sorry, I didn't look at this first). I tried this example with and without the limit, and it returned the same number of rows, which is what I would've expected with the bbox query. Strange behavior that the example above works with pages of 100 but not pages of 10,000?

R/read_USGS_daily.R
ldecicco-USGS: What I'm guessing... it seems like we can get 20 pages at most. At least in my little bit of testing, I'm seeing 20 pages coming back without a warning that you are missing data. I'll try to confirm, but here: what I see:

```r
bbox_lim_100 <- read_USGS_daily(bbox = c(-83, 36.5, -81, 38.5), 
                          parameter_code = c("00060", "00065"), 
                          limit = 100)
nrow(bbox_lim_100)
2000

bbox_lim_500 <- read_USGS_daily(bbox = c(-83, 36.5, -81, 38.5), 
                          parameter_code = c("00060", "00065"), 
                          limit = 500)
nrow(bbox_lim_500)
10000

bbox_lim_1000 <- read_USGS_daily(bbox = c(-83, 36.5, -81, 38.5), 
                          parameter_code = c("00060", "00065"), 
                          limit = 1000)
nrow(bbox_lim_1000)
20000

bbox_lim_2000 <- read_USGS_daily(bbox = c(-83, 36.5, -81, 38.5), 
                          parameter_code = c("00060", "00065"), 
                          limit = 2000)
nrow(bbox_lim_2000)
40000

```

PR #774, opened 2025-05-21
don't assume order of nldi sources -- turn nldi tests back on but off in CI

dblodgett-usgs:
We have a new version of the NLDI coming and this is the only issue I see where the new one doesn't work with dataRetrieval. This change is backward compatible.

### DIFF
--- a/R/findNLDI.R
+++ b/R/findNLDI.R
@@ -211,7 +211,8 @@ valid_ask <- function(all, type) {
 
   ### WOW! This is hacky and will hopefully be unneeded latter on....
   type <- ifelse(type == "nwis", "nwissite", type)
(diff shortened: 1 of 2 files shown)


ldecicco-USGS:
So....do we add dplyr as a import for this change? I went through some effort many years ago to get rid of it as a dependency because a few squeeky wheels were so anti-dplyr at the time, and it wasn't adding _much_. We already Suggest it, so it's not the end of the world.

dblodgett-usgs:
Oh -- do you know import dplyr? I had no idea. I'll implement an alternate solution. Sorry about that.

dblodgett-usgs:
Looks like I don't actually need bind_rows().

PR #777, opened 2025-06-02
Add documentation for internal functions

ldecicco-USGS:


ldecicco-USGS:
@jzemmels , @ehinman, @cnell-usgs

This is:
1. Notes in the form of roxygen for internal functions. There are different schools of thought on how much to document and test internal functions. On one hand, we don't want to promise users that those functions will be maintained (so that is the camp that says don't test or create Rd files). On the other hand, we need some help remembering what was suppose to happen (so that's the camp that recommends @keyword internal for docs and shoot for 100% code coverage). I think this is a decent middle ground - each function has some example code to help us troubleshoot stuff down the road, but the @noRd tag assures those pages are never rendered. The noRd is what rOpenSci recommends
2. Moves Status page to a more prominent location in the pkgdown header.
3. Added a message to all NWIS functions:

```r
ALERT: All NWIS services are slated for decommission.
For up-to-date information, see: 
https://doi-usgs.github.io/dataRetrieval/articles/Status.html
```
4. Bonus: found some tests I missed converting from the NWIS functions to USGS.

PR #782, opened 2025-06-05
Fix ids

ldecicco-USGS:
1. Add max_results
2. Add some logic that a user could request monitoring_location_id or monitoring_locations_id (plural) to get the "id" column from the monitoring-locations service. 
3. Add same logic that the "id" column from the time-series-metadata service actually comes back as "timeseries_id" since that is the name that other functions will require.
4. Added some "lessons learned" in the vignette
5. Changed the way the data is parsed if skipGeometry=TRUE. The way it was coding was dropping the "id" column, now it uses sf::read_sf even if skipGeometry=TRUE, but then drops the empty geometry.

### DIFF
--- a/R/construct_api_requests.R
+++ b/R/construct_api_requests.R
@@ -43,7 +43,8 @@
 construct_api_requests <- function(service,
                                    properties = NA_character_,
                                    bbox = NA,
-                                   limit = 10000,
+                                   limit = NA,
+                                   max_results = NA,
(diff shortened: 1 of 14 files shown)


ldecicco-USGS:
Not sure what's going on with GitHub Actions. It was hanging forever and eventually cancelled itself. The tests ran and passed locally and on Gitlab, so for now let's assume things pass the checks/tests

PR #803, opened 2025-07-11
Metadata collections

ldecicco-USGS:


ldecicco-USGS:
@ehinman @jzemmels @cnell-usgs here are the metadata functions that I'd like to bundle in the next CRAN update (due July  20). If you could pull down these changes and give them a try, that would be great.

jzemmels (approved):
New functions look great. I've ran the examples you've provided and looked over the documentation.

There are a few suggested changes that can be incorporated in the future, if it all, as I know you want to get this to CRAN ASAP. Approved.

PR #845, opened 2026-01-17
NLDI client update fixes #844

### LINKED ISSUE #844: NLDI nwissite now includes agency code prefix on site ids.
**Describe the bug**

I've migrated the NLDI to use agency code prefixes and should no longer be stripping the "USGS-" off the site ides. I'll open a PR in a moment with the change.

dblodgett-usgs:
I've been working to get the agency code IDs into the overall NLDI client and these changes are needed for the way I call this code from nhdplusTools. No change to the tests here and generally relaxes some of the "special" handling that was in here.

### DIFF
--- a/R/findNLDI.R
+++ b/R/findNLDI.R
@@ -177,25 +177,6 @@ get_nldi <- function(url, type = "", use_sf = FALSE, warn = TRUE) {
   }
 }
 
-#' Clean NWIS NLDI ids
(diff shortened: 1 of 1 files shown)


ldecicco-USGS:
We're trying to stage all changes that will eventually be sent to CRAN in the develop branch. When I'm about ready to submit, I'll make a PR for develop -> main to verify everything looks good. That should actually be this week because we want to get "continuous" out there asap.

PR #866, opened 2026-03-02
Add wildcard CQL

ldecicco-USGS:
This PR includes some minor fixes for the field and combined meta. 

It switches the WQP table reader to data.table::fread. There are a few minor differences in the WQX3 table, but the fread conversion looks closer to the actual returned text.

The big functional change is adding a wildcard CQL2 template which will allow this:

```r
hucs <- read_waterdata_combined_meta(
+   hydrologic_unit_code = c("11010008", "11010009"),
+   site_type = c("Stream", "Spring")
+ )
Requesting:
https://api.waterdata.usgs.gov/ogcapi/v0/collections/combined-metadata/items?f=json&lang=en-US&limit=50000
Remaining requests this hour:2275 
                                   
> unique(hucs$site_type)
[1] "Stream" "Spring"
> unique(hucs$hydrologic_unit_code)
 [1] "110100080704" "110100090104" "110100090202" "110100080302" "110100080203"
 [6] "11010008"     "110100080212" "110100080102" "110100080802" "110100090107"
[11] "110100080904" "110100080603" "110100090306" "110100080502" "110100080701"
[16] "110100080209" "110100080902" "110100080206" "110100080309" "110100080210"
[21] "110100090201" "110100090204" "110100080111" "110100080801" "11010009"    
[26] "110100080211" "110100080401" "110100080104" "110100080308" "110100080903"
[31] "110100080806" "110100080101" "110100090302" "110100090305" "110100080112"
[36] "110100090206" "110100090103"
```

So the logic is if a user specifies >1 HUC, we'll use the wildcard "like" and "or" template (Huc 0123% OR Huc01222%). Any other parameters will be tacked on with ANDs

### DIFF
--- a/vignettes/daily_data_statistics.Rmd
+++ b/vignettes/daily_data_statistics.Rmd
@@ -40,7 +40,7 @@ site1 <- "USGS-02037500"
 
 Day-of-year and month-of-year statistics aggregate observations for the same calendar day or month across multiple years to describe typical seasonal conditions (e.g., all Januarys or all January 1sts).
 Consider the output below, where we request day-of-year discharge averages for January 1 and January 2.
-Note that the `start_date` and `end_date` are set in `month-year` format to describe the day-of-year range.
+Note that the `start_date` and `end_date` are set in `month-day` format to describe the day-of-year range.
 
 ```{r}
 jan_por_mean <-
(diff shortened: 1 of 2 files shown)


R/importWQP.R
```diff
@@ -62,14 +62,12 @@ importWQP <- function(obs_url, tz = "UTC",
...
-  
-  if(last_chars != c("\n")){
-    doc <- paste0(doc, "\n")
-  }
-  retval <- suppressWarnings(readr::read_delim(doc,
-                                               col_types = readr::cols(.default = "c"),
-                                               quote = ifelse(csv, '\"', ""),
-                                               delim = ifelse(csv, ",", "\t")))
+
+  retval <- data.table::fread(text = doc, data.table = FALSE,
```
jzemmels: 😎

vignettes/daily_data_statistics.Rmd
```diff
@@ -35,7 +42,7 @@ Day-of-year and month-of-year statistics aggregate observations for the same cal
 Consider the output below, where we request day-of-year discharge averages for January 1 and January 2.
 Note that the `start_date` and `end_date` are set in `month-year` format to describe the day-of-year range.
```
jzemmels: My typo
```suggestion
Note that the `start_date` and `end_date` are set in `month-day` format to describe the day-of-year range.
```

ldecicco-USGS:
> 
> In the future, exposing the templates more explicitly might be a good, flexible way for users to make more sophisticated queries. Perhaps the format could be building a CQL query string piece-by-piece using a combination of the template helper functions. From a set algebra perspective, I think the primitive operations are `and`, `or`, and `not`.

We can keep this on the backburner. Every time I start going down a thought process like this, I don't come up with a way that is significantly easier for users compared to giving some CQL2 examples (especially since I think we've got the vast majority of user's needs covered with current functionality).

PR #878, opened 2026-04-13
Fix bugs, remove dead code, and deduplicate shared patterns

thodson-usgs:
## Summary
- Fix `format_api_dates` bug where loop variable `i` was unused — full datetime vector passed to `get_dateTime()` instead of each element
- Fix `readNWISrating` referencing undefined `intColumns` variable (would crash at runtime)
- Remove dead functions (`only_legacy`, `post_url`, `.capitalALL`) and unreachable `measurements` switch branch
- Extract `add_api_token()` helper replacing 4 duplicate token injection patterns
- Extract `coerce_num_cols()` and `log_rate_limit()` helpers in `walk_pages.R`
- Simplify deprecated `read_USGS_samples` wrapper from 60 lines to 5 using `...` forwarding
- Cache user-agent string in `pkg.env` to avoid recomputing on every HTTP request
- Replace O(n²) rbind-in-loop with pre-allocated lists in `readNGWMNdata`

Net: -131 lines across 10 files. All 283 tests pass.

## Test plan
- [x] All 283 existing unit tests pass (`devtools::test()`)
- [ ] Verify `read_waterdata_samples` deprecated alias still works
- [ ] Verify datetime range queries produce correct ISO8601 output
- [ ] Verify `readNWISrating` works when `current_rating_nu` column is present


ldecicco-USGS:
Could you change the PR to aim to the "develop" branch? We try to keep "main" to what was most recently pushed to CRAN, and "develop" to what will eventually get pushed up there. The main reason for that (vs just a tagged release) is making it easier to keep the GH-Actions that produce the documentation sync'ed up to what's on CRAN and not a mis-match with what _will_ be on CRAN (since the vast vast majority of users get the package from there and not GitHub). 

I think the datetime bug was already fixed on develop. Some of the other shared pattern stuff looks good.

thodson-usgs:
@ldecicco-USGS rebased to dev. Please leave comments on any changes you don't want, or close this PR entirely. I was mainly curious what Claude wouild drag up.

PR #879, opened 2026-04-16
Move chunk_size to argument and retry_on_failure = TRUE

ldecicco-USGS:
After some testing, realized the "req_retry" function wasn't really doing much unless `retry_on_failure = TRUE` was set and `httr2::req_perform_iterative` needs to have `on_error = "return"` so that the retry actually gets triggered.

Additionally, moved the chunking argument to a top level argument.


ldecicco-USGS:
Also a lot of argument rejiggering to follow tidyverse design principles:
https://design.tidyverse.org/dots-after-required.html

R/construct_api_requests.R
```diff
@@ -74,9 +107,10 @@ construct_api_requests <- function(
...
     "time",
-    "limit",
     "begin_utc",
-    "end_utc"
+    "end_utc",
+    "limit",
+    "skipGeometry"
   )
 
   comma_params <- c(
```
jzemmels, line 116: Unrelated but a question: do you know if `hydrologic_unit_codes` from the monitoring-location endpoint is a single or comma parameter? I was just testing out passing a list of HUC8s in drpy, and I got an answer, but the count was way small and I don't even know what was actually used.

R/read_waterdata_parameter_codes.R
```diff
@@ -18,11 +18,16 @@
 #' Available options are:
 #' `r dataRetrieval:::get_properties_for_docs("parameter-codes", "parameter_code_id")`.
 #' The default (`NA`) will return all columns of the data.
+#' @param \dots Not used. Included to help differentiate official Water Data API arguments
```
jzemmels, line 21: I'm not sure what this means. Do you mean differentiate within the documentation file? Like a visual boundary between the API vs. dR-specific parameters?

R/get_ogc_data.R
```diff
@@ -179,3 +169,103 @@ switch_properties_id <- function(properties, id) {
...
+#' depth). Coordinates are assumed to be in crs 4326. The expected format is a numeric
+#' vector structured: c(xmin,ymin,xmax,ymax).
+#' Another way to think of it is c(Western-most longitude,
+#' Southern-most latitude, Eastern-most longitude, Northern-most longitude).
+#' @param skipGeometry This parameter can be used to skip response geometries for
+#' each feature. The returning object will be a data frame with no spatial
+#' information. The default `NA` will not specify the argument in the request.
+#'
+#' @keywords internal
+check_arguments_api <- function(bbox, skipGeometry, ...) {
```
jzemmels: What are the dots used for here?

R/read_waterdata_ratings.R
```diff
@@ -0,0 +1,168 @@
...
+#'
+#' bbox_query <- read_waterdata_ratings(bbox = bbox,
+#'                                      download_and_parse = FALSE)
+#' length(bbox_query)
+#' recent_query <- read_waterdata_ratings(bbox = bbox,
+#'                                        datetime = c(Sys.Date()-7, NA),
+#'                                        download_and_parse = FALSE)
+#' length(recent_query)
+#'}
+read_waterdata_ratings <- function(
```
jzemmels, line 77: I accidentally ran this while at the API limit and got back the following output:

```r
trying URL '[https://api.waterdata.usgs.gov/stac-files/ratings/USGS.01104475.exsa.rdb](vscode-file://vscode-app/c:/Users/jzemmels/AppData/Local/Programs/Positron/resources/app/out/vs/code/electron-browser/workbench/workbench.html#)'
Warning message:
In utils::download.file(url = url, destfile = full_file_path) :
  cannot open URL '[https://api.waterdata.usgs.gov/stac-files/ratings/USGS.01104475.exsa.rdb](vscode-file://vscode-app/c:/Users/jzemmels/AppData/Local/Programs/Positron/resources/app/out/vs/code/electron-browser/workbench/workbench.html#)': HTTP status was '429 Unknown Error'
```

Do we want the retry and print functionality to extend to this function as well?

R/read_waterdata_ratings.R
```diff
@@ -0,0 +1,168 @@
...
+#' @param \dots Not used.
+#' @param limit Limits the number of results that are included in each page of
+#' the response (capped at the default 10,000).
+#' @param download_and_parse Logical to define whether or not to download, parse,
+#' and return a list of data frames with rating curve data (`TRUE`), or to return
+#' just a list of available rating curve files (`FALSE`). Default is `TRUE`.
+#' @export
+#' @inherit read_waterdata_continuous details
+#'
+#' @return List of data frames which contain the requested rating curves.
```
jzemmels, line 41: Is a list and those column names the expected output from users? Do you think users would prefer a row-bound, "tidy" data frame with an extra column indicating the site ID? I'm not saying I'd prefer the tidy format, just wondering what a user would expect.

R/read_waterdata_ratings.R
```diff
@@ -0,0 +1,168 @@
...
+
+download_convert <- function(feature, file_path, file_type) {
+  links <- feature$links
+  id <- feature$id
+  url <- feature$assets$data$href
+
+  if (any(sapply(file_type, function(x) grepl(x, url)))) {
+    full_file_path <- file.path(file_path, id)
+    utils::download.file(url = url, destfile = full_file_path)
+    rating <- dataRetrieval::importRDB1(full_file_path)
```
jzemmels:
```suggestion
    rating <- importRDB1(full_file_path)
```

jzemmels (commented):
These updates look good. I have a few comments and questions, but nothing major. I ran check() locally and didn't see anything worrisome.

R/read_waterdata_parameter_codes.R
ldecicco-USGS: Riffing off tidyverse's design principle:
https://design.tidyverse.org/dots-after-required.html

The idea is that the arguments after the ...'s are something rarely used and this requires that the users would specifically name them. Granted, in our functions, users mostly need to name the args anyway since there are so many (and no one should be defining ALL the arguments), but essentially the answer is yes - it's a good visual way to differentiate what are queries used in the API, and what we use outside the API (and therefore what users should either not mess with too often, and if they do they should understand what they're doing)

I also put "limit" in the dR-specific because it falls more into the category that users should rarely need to mess with it and it's confusing. They might think if they set the limit=10 they will only get 10 results, but it actually means they'll get 10 per page....so in that regard, since dR does automatic page walking.

R/construct_api_requests.R
ldecicco-USGS: When I last asked about that, huc's were not going to be comma parameters "any time soon" (so maybe someday?).

Remember HUCs are tricky. Our sites have a many combination of number of digits. So I think the issue was if you asked for c("012345", "012346"), you would only get those 2 HUCs, even if there were tons of sites with "01234567" and "01234678" for instance.

In dataRetrieval we send the HUC requests to a POST and that allows us to take advantage of the work the API developers put in to make HUCs special so that it really is going is as c("012345%", "012346%")

R/read_waterdata_parameter_codes.R
jzemmels: Ahh, I understand. Thanks for clarifying

R/get_ogc_data.R
ldecicco-USGS: nothing, pulling them out now...

R/read_waterdata_ratings.R
ldecicco-USGS: good catch! I'll see if it acts the same way in the header as the ogc stuff

R/read_waterdata_ratings.R
ldecicco-USGS: The issue is that there is a lot of important information hidden in the "comment" attribute. So somehow we'd need to bundle that information in as well. On the other hand, the list means that each data frame acts the same as the traditional 1 site returned from `readNWISrating`.

R/read_waterdata_ratings.R
ldecicco-USGS: Well, the headers look the same for the request, but the download.file is outside httr2, so it won't be trivial to use something like the `httr2::req_error(body = error_body)` that we use with the APIs

ldecicco-USGS: Nevermind, got the download working with httr2 instead of the base download.file.

R/construct_api_requests.R
ldecicco-USGS: Oh wait! I forgot I set up the huc POST stuff specifically:
https://github.com/DOI-USGS/dataRetrieval/blob/develop/R/construct_api_requests.R#L498

PR #888, opened 2026-05-01
Peaks and Ratings

ldecicco-USGS:
Also commenting out any NWIS or WQP examples. 

Removed moving average vignette because HASP is superior.

Removed old stat vignette because new stat service is superior (we were thinking about keeping it up because it shows how to make maps but we have that in the waterdata API article and tutorials).


R/construct_api_requests.R
```diff
@@ -370,38 +370,57 @@ switch_arg_id <- function(ls, id_name, service) {
...
 #' end <- Sys.Date()
 #' dataRetrieval:::format_api_dates(c(start, end), date = TRUE)
+#'
+#' # This is a problem because the first value forces the
+#' # vector to be numeric, and then we don't really
+#' # know if the 2nd value is a Date (number of days since 1970)
+#' # or if it's a date/time (number of seconds..)
+#' half_range <- c(NA, as.Date("2025-01-01"))
+#' # Will error:
+#' #dataRetrieval:::format_api_dates(half_range, date = FALSE)
```
jzemmels, line 380: This didn't error for me, but it did appear to interpret as a numeric timestamp, meaning I got `"../1970-01-01T05:34:49Z"` as output

R/construct_api_requests.R
```diff
@@ -370,38 +370,57 @@ switch_arg_id <- function(ls, id_name, service) {
...
   if (length(datetime) > 2) {
     stop("datetime should only include 1-2 values")
   }
 
+  if (is.numeric(datetime)) {
+    # Until we can figure out a way to know if the
+    # original input was suppose to be Date or Posix
+    # We can't determine what the user meant.
+    stop(
+      "A time query was entered as numeric. This could lead to errors.
```
jzemmels, line 403: The only way I can think to do this would be to check the expression passed within the function call, but that gets into `substitute` and related territory that I know we want to avoid. Otherwise, the function could change to accept separate `date_start`/`end` arguments or a list instead of a vector. I can't think of other ways.

R/read_waterdata_peaks.R
```diff
@@ -0,0 +1,81 @@
...
+#'                data_type = "Peaks",
+#'                parameter_code = "00060")
+#'
+#'
+#' dv_data_sf <- read_waterdata_peaks(
+#'                monitoring_location_id = wi_peaks$monitoring_location_id[1],
+#'                parameter_code = "00060")
+#'
+#' }
+read_waterdata_peaks <- function(
```
jzemmels, line 49: One thing I recently noticed about the combined_metadata endpoint is that it does not return a `time_series_id`, only a `parent_time_series_id`. It's possible that peaks don't have a dedicated timeseries ID, but I'm also seeing that the parent timeseries ID are all `NA` in the Wisconsin site example: https://api.waterdata.usgs.gov/ogcapi/v0/collections/combined-metadata/items?f=json&lang=en-US&parameter_code=00060&data_type=Peaks&state_name=Wisconsin&limit=50000

I would expect there to at least be a parent time series ID of daily values or something. But I've also noticed that combined_metadata doesn't return the TS ID if fetching daily values either. I don't think this is a dataRetrieval problem, but is perhaps an unexpected behavior, at least as I understand the endpoint? I would expect that the data returned by `timeseries_metadata` would be a subset of columns returned by `combined_metadata` but that doesn't appear the case.

tutorials/quick_intro_deck.qmd
```diff
@@ -147,11 +148,18 @@ library(dataRetrieval)
 
 ### Python
 
+Whether you are a user or developer we recommend installing `dataretrieval` in a virtual environment. This can be done using something like virtualenv or conda.
```
jzemmels:
```suggestion
Whether you are a user or developer we recommend installing `dataretrieval` in a virtual environment. This can be done using something like [virtualenv](https://docs.python.org/3/tutorial/venv.html) or [conda](https://docs.conda.io/projects/conda/en/stable/user-guide/install/index.html). 
```

tutorials/basic_slides_deck.qmd
jzemmels, line 363: Do we want to add the parallel Python workflow in these slides? We can add that later too

tutorials/basic_slides_deck.qmd
```diff
@@ -167,6 +186,33 @@ Then each time you open R, you'll need to load the library:
...
+#| eval: false
+pip install dataretrieval
+```
+
+or
+
+```{bash}
+#| echo: true
+#| eval: false
+conda -c conda-forge install dataretrieval
```
jzemmels, line 204: Just a note for rendering these slides specifically: I had to do some reticulate stuff to set up a conda env properly. Namely, creating a conda env then setting the `RETICULATE_PYTHON` environment variable . Not sure how you've set it up locally but it may be good to put somewhere how to render these specifically?

tutorials/basic_slides_deck.qmd
```diff
@@ -447,25 +507,52 @@ We're going walk through 3 retrievals:
...
+
+```{python}
+#| eval: !expr evaluate_python
+from dataretrieval import waterdata
+
+site = "USGS-09405500"
+pcode = "00060"  # Discharge
+stat_cd = "00003"  # Mean
 
+df, md = waterdata.get_daily(
```
jzemmels, line 544: Worth noting (somewhere) that the drpy prints a message if `geopandas` isn't installed about flattening the geometry column. I'm guessing it won't matter for most, but I anticipate new users may be confused by the message.

tutorials/basic_slides_deck.qmd
jzemmels, line 649: Just a note: this text is low on the slide and ends up getting cropped in my local version. I see it's not an issue in the website's version so I think it's fine.

tutorials/basic_slides_deck.qmd
```diff
@@ -564,14 +680,16 @@ The "time" argument has a few options:
 Here are a bunch of valid inputs:
 
 ```{r}
-#| code-line-numbers: "1-7|8-9|10-13|14-17"
+#| code-line-numbers: "1-9|10-11|12-15|16-19"
```
jzemmels, line 683: Love the code highlighting

tutorials/quick_intro_deck.qmd
jzemmels: Is this just to highlight the columns returned by the function? If yes, could that be added as text to the slide?

The drpy function returns different columns in a different order. 

``` py
>>> continuous_data.columns
Index(['time_series_id', 'monitoring_location_id', 'parameter_code',
       'statistic_id', 'time', 'value', 'unit_of_measure', 'approval_status',
       'qualifier', 'last_modified', 'continuous_id'],
      dtype='str')
```

vignettes/Status.Rmd
jzemmels, line 1: I'm not seeing this updated in the Status article on the package website. Have the webpages updated?

tutorials/basic_slides_deck.qmd
jzemmels, line 1: Now that the number of chunks are almost doubling, it could be helpful to add chunk labels for troubleshooting. For example:

`#| label: api-pat-r`
`#| label: api-pat-py`

jzemmels (approved):
Changes look good. I ran the new peaks examples, re-ran unit tests, and rendered the slidedecks. All ran without issues. Just a few comments.

R/construct_api_requests.R
ldecicco-USGS: I wonder if I pushed the "stop" command up after you pulled? I get:

```r
> half_range <- c(NA, as.Date("2025-01-01"))
> dataRetrieval:::format_api_dates(half_range, date = FALSE)
Error in dataRetrieval:::format_api_dates(half_range, date = FALSE) : 
  A time query was entered as numeric. This could lead to errors.
Check any time queries that might have been automatically converted to numeric.
This could happen if you use c(NA, as.Date(Sys.Date())) instead of
as.Date(c(NA, Sys.Date()) for example.
```

R/construct_api_requests.R
ldecicco-USGS: I think most workflows I've seen are using that use characters, Date objects, or POSIX. All of those are fine, and any numeric is unintentional. So I'm comfortable leaving this as an error. I assume the time it will come up most is something like:

```r
df <- read_waterdata_sample(some stuff)
df_range <- range(df$time)
cont_df <- read_waterdata_daily(time = c(NA, df_range[2])
```
If we plaster examples and tutorials with stuff like:
```r
cont_df <- read_waterdata_continuous(time = as.Date(c(NA, df_range[2]))
```
I think we'll be OK.

tutorials/basic_slides_deck.qmd
ldecicco-USGS: We do, and yeah I expect another PR with a bunch of updated Python stuff eventually

tutorials/basic_slides_deck.qmd
ldecicco-USGS: That's in here:
https://water.code-pages.usgs.gov/dataRetrieval/articles/Contributing.html#setting-up-quarto-slides
(which I think is part of this PR in Contributing.Rmd) Let me know if you think anything there needs changing

tutorials/basic_slides_deck.qmd
ldecicco-USGS: I _assumed_ geopandas would bring us over the allowed Docker space which is why I haven't added it to the environment, but I don't remember if I tested that specifically or not. I'm sure it's OK for GHActions. Maybe we can make the Gitlab ci only install the python packages in the tutorial render step...I'll give that a try then folks wouldn't see the message.

tutorials/basic_slides_deck.qmd
ldecicco-USGS: The `air` auto-formatting screwed all these up. Not that I'm complaining, it looks way better now.

tutorials/basic_slides_deck.qmd
ldecicco-USGS: You're right but ugg... maybe it _is_ time to learn AI 😂🤖

vignettes/Status.Rmd
ldecicco-USGS: https://water.code-pages.usgs.gov/dataRetrieval/articles/Status.html
all development updates are on code.usgs.gov, the github pages don't change until we merge things into main. Assuming that's what you meant (?)

R/construct_api_requests.R
jzemmels: That sounds like a good plan.

tutorials/quick_intro_deck.qmd
ldecicco-USGS: There's no super slick way in revealJS slides to have the output from a code chunk be in a column on the side that is NOT an even 50:50 split. If it's even sized columns, this particular slide looks bad. So that's the original reason I hard-coded the column names.

But for completeness, I added the ol' classic trick of one side is eval: false, the other side is echo:false.

R/read_waterdata_peaks.R
ldecicco-USGS: It returns a column combined_meta_id, which is either the time_series_id, or the field_series_id:

```r
site <- "USGS-05407000"
available_combine <- read_waterdata_combined_meta(monitoring_location_id = site)
avail_ts <- read_waterdata_ts_meta(monitoring_location_id = site)
avail_field <- read_waterdata_field_meta(monitoring_location_id = site)

all(avail_ts$time_series_id %in% available_combine$combined_meta_id)
[1] TRUE
all(avail_field$field_series_id %in% available_combine$combined_meta_id)
[1] TRUE
```

I'll look into what the combined_meta_id means for Peaks, because I really don't know, it doesn't seem to match anything. Also...I'll try to describe that better in the docs...

ldecicco-USGS: Actually, the combined_meta_id (or time_series_id from the ts_meta function) can be the input time_series_id in `read_waterdata_peaks`. Not sure if a parent_time_series can be tied back to peaks.

ldecicco-USGS: This made me realize the field_meta id should be "field_measurements_series_id" and not "field_series_id" to be consistent with the `read_waterdata_field_measurements` function. I'm changing it now so it doesn't confuse us for all time.

PR #890, opened 2026-05-11
Add normal_type and interval_type arguments to stats functions

jzemmels:
Two new arguments to the stats functions:

- `normal_type` for `_stats_por` function accepts values `"DOY"`, `"MOY"`.
- `interval_type` for `_stats_daterange` function accepts values `"M"`, `"CY"`, `"WY"`.

Updated function docs to reflect these updates.

The API also returns two new columns that aren't filterable ahead of time: `parent_statistics_id` and `parent_statistics_name`. I don't think these require an update.

### DIFF
--- a/R/read_waterdata_stats.R
+++ b/R/read_waterdata_stats.R
@@ -39,6 +39,13 @@
 #'   supplied then statistics will be supplied for the entire period of record.
 #' @param end_date End Date Query Parameter. The logic is inclusive i.e., it will
 #'   also return records that match the date.
+#' @param normal_type Normal Type Query Parameter. If unspecified, all matching data
+#'   will be returned. Otherwise, it will filter the results to one of the following
+#'   normals: day-of-year, month-of-year. Available values: "DOY", "MOY".
+#' @param interval_type Interval Type Query Parameter. If unspecified, all matching
+#'   data will be returned. Otherwise, it will filter the results to one or more of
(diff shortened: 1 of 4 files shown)


R/read_waterdata_stats.R
```diff
@@ -88,6 +101,12 @@
 #'   monitoring_location_id = c("USGS-02319394", "USGS-02171500")
 #' )
 #'
+#' # Request only calendary year statistics
```
ldecicco-USGS: calendary -> calendar

ldecicco-USGS (approved):
Looks good other than the 1 spelling (and my spelling is atrocious so maybe that's a word I don't know...)

R/read_waterdata_stats.R
jzemmels: That's a typo. Thanks for catching!

ldecicco-USGS:
Looks like you need to run `devtools::document()` (or whatever your favorite way to update the roxygen tags) and then push up the changes to the .Rd file for the check to pass.

jzemmels:
> Looks like you need to run `devtools::document()` (or whatever your favorite way to update the roxygen tags) and then push up the changes to the .Rd file for the check to pass.

Ah, yep. Thanks for the reminder!

jzemmels:
Looks like the checks have passed now.

PR #893, opened 2026-05-14
nldi client tweaks

dblodgett-usgs:
The NLDI provides a mainstem ID that we need to add, and this also makes it so you can switch to a different endpoint with options -- I think I followed conventions from the rest of the package but am happy to switch things up if we need.


ldecicco-USGS:
I'll fix that warning with a PR I'm about to push

PR #896, opened 2026-05-20
Force geometry to WGS84 if missing, handle missing values

jzemmels:
Force CRS to be WGS84 if it's missing (can happen if API response is empty)

Also, some extra logic for handling empty `values` returned by stats API (seems to be a periodic clerical error in the database).


ldecicco-USGS (approved):
Looks good!

PR #904, opened 2026-06-11
Ngwmn

ldecicco-USGS:
Initial PR for adding the new NGWMN functions:

https://api.waterdata.usgs.gov/ngwmn/ogcapi/collections


R/get_ogc_documentation.R
```diff
@@ -55,6 +59,7 @@ get_collection <- function() {
 #' available, it will also contain a description.
 #'
 #' @param service Character, can be any of the USGS Waterdata API endpoints or collections.
+#' @param base Either "OGC" for waterdata, or "NGWMN".
```
ehinman:
```suggestion
#' @param base Either "OGC" for waterdata, or "NGWMN" for National Groundwater Monitoring Network.
```

R/read_ngwmn_lithology.R
```diff
@@ -0,0 +1,52 @@
...
+read_ngwmn_lithology <- function(
+  monitoring_location_id = NA_character_,
+  monitoring_location_obs_number = NA_character_,
+  properties = NA_character_,
+  ...,
+  convertType = getOption("dataRetrieval.convertType"),
+  no_paging = getOption("dataRetrieval.no_paging"),
+  chunk_size = getOption("dataRetrieval.site_chunk_size_data"),
+  limit = getOption("dataRetrieval.limit"),
+  attach_request = getOption("dataRetrieval.attach_request")
```
ehinman, line 38: These make it so that you don't have to retype the exact same default over and over?

R/read_ngwmn_sites.R
```diff
@@ -0,0 +1,131 @@
...
+#' The default (`NA`) will return all columns of the data.
+#' @param skipGeometry This option can be used to skip response geometries for
+#' each feature. The returning object will be a data frame with no spatial
+#' information.
+
+#' @param bbox Only features that have a geometry that intersects the bounding
+#' box are selected.The bounding box is provided as four or six numbers, depending
+#' on whether the coordinate reference system includes a vertical axis (height or
+#' depth). Coordinates are assumed to be in crs 4326. The expected format is a numeric
+#' vector structured: c(xmin,ymin,xmax,ymax). Another way to think of it is c(Western-most longitude,
```
ehinman, line 69: If I wanted to include a vertical axis with height and depth, what would that look like in a vector?

R/read_ngwmn_lithology.R
ldecicco-USGS: yeah, you could either put this in a script or in your .Renviorn file:

```r
options(dataRetrieval.attach_request = FALSE)
```

If it's in something like the .Renviorn file, it'll stay that way until you change it (so if you really don't like the request attached as an attribute)

R/read_ngwmn_well_construction.R
```diff
@@ -0,0 +1,56 @@
+#' Get NGWMN Well Construction Data
+#'
+#' @description `r get_description("constructionObs", base = "NGWMN")`
+#'
+#' @export
+#' @param monitoring_location_id
```
ehinman, line 6: With the exceptions of the sites and providers services, the swagger doc says that monitoring location id is required: https://api.waterdata.usgs.gov/ngwmn/ogcapi/openapi?f=html#/waterLevelObs

Is that noted anywhere?

ehinman: Maybe it's obvious.

R/read_ngwmn_well_construction.R
```diff
@@ -0,0 +1,56 @@
...
+#' `r get_ogc_params("constructionObs", base = "NGWMN")$monitoring_location_id$description`
+#' @param monitoring_location_obs_number
```
ldecicco-USGS, line 8: When this line is rendered it says :

"This field is required. Combined site identifier of agency code and site number (format of `{agency_code}`-`{monitoring_location_number}`). A list of values can be passed for this field, seperated by commas.\n"

Admittedly I hadn't noticed that so we can put in a check that the argument it's NA

R/setAccess.R
```diff
@@ -77,5 +77,6 @@ Please contact comptools@usgs.gov for more information."
   pkg.env$status <- "https://www.waterqualitydata.us/wqx3/status/"
 
   pkg.env$NGWMN <- "https://cida.usgs.gov/ngwmn_cache/sos"
+  # pkg.env$NGWMN <- "https://www.usgs.gov/apps/ngwmn/ngwmn_cache/sos"
```
ehinman: Needed?

R/read_ngwmn_lithology.R
```diff
@@ -0,0 +1,52 @@
...
+#' @export
+#' @param monitoring_location_id
+#' `r get_ogc_params("lithologyObs", base = "NGWMN")$monitoring_location_id$description`
+#' @param monitoring_location_obs_number
+#' `r get_ogc_params("lithologyObs", base = "NGWMN")$monitoring_location_obs_number$description`
+#' @param properties A vector of requested columns to be returned from the query.
+#' Available options are:
+#' `r dataRetrieval:::get_properties_for_docs("lithologyObs", base = "NGWMN")`.
+#' The default (`NA`) will return all columns of the data.
+#' @inheritParams check_arguments_non_api
```
ehinman, line 14: Minor, but it looks like `dots` is included and the description is "not used". Then why include as an argument?

R/read_ngwmn_providers.R
```diff
@@ -0,0 +1,55 @@
...
+#'
+#' @examplesIf is_dataRetrieval_user()
+#'
+#' \donttest{
+#'
+#' ngwmn_providers <- read_ngwmn_providers(state = "WI")
+#'
+#' ngwml_providers2 <- read_ngwmn_providers(state = c("WI", "MN"))
+#'
+#' org_type <- read_ngwmn_providers(organization_type = "NWIS", state = c("WI", "MN"))
```
ehinman, line 26: The link column shows up as `""` if there is no associated link. Seems fine, but I first wondered if all of them were empty and saw a few in CA with actual URLs pointing to provider webpages.

R/read_ngwmn_sites.R
```diff
@@ -0,0 +1,131 @@
...
+#' @description `r get_description("sites", base = "NGWMN")`
+#'
+#' @export
+#' @param monitoring_location_id
+#' `r get_ogc_params("sites", base = "NGWMN")$monitoring_location_id$description`
+#' @param agency_code
+#' `r get_ogc_params("sites", base = "NGWMN")$agency_code$description`
+#' @param monitoring_location_number
+#' `r get_ogc_params("sites", base = "NGWMN")$monitoring_location_number$description`
+#' @param altitude
```
ehinman, line 12: Future enhancement could turn this into a range input, since right now you just give it a number as a character. I tried `test <- read_ngwmn_sites(altitude = "100")` and `test <- read_ngwmn_sites(altitude = "100.25")` just for fun and the first returned a few sites. But then looking at something like `test <- read_ngwmn_sites(state_name = "Minnesota")`, altitudes do go out to the hundredths place and are very specific, so unless you know the exact altitude(s) you want, this input seems pretty useless.

R/read_ngwmn_water_level.R
```diff
@@ -0,0 +1,82 @@
...
+#' @param sample_time
+#' `r get_ogc_params("waterLevelObs", base = "NGWMN")$sample_time$description`
+#' See also Details below for more information.
+#'
+#' @param data_provided_by
+#' `r get_ogc_params("waterLevelObs", base = "NGWMN")$data_provided_by$description`
+#' @param water_depth_below_land_surface_ft
+#' `r get_ogc_params("waterLevelObs", base = "NGWMN")$water_depth_below_land_surface_ft$description`
+#' @param water_level_above_site_datum_ft
+#' `r get_ogc_params("waterLevelObs", base = "NGWMN")$water_level_above_site_datum_ft$description`
```
ehinman, line 19: Same comment as above about someday making these ranges, rather than singular strings.

R/read_ngwmn_water_level.R
```diff
@@ -0,0 +1,82 @@
...
+#' @param monitoring_location_vertical_datum
+#' `r get_ogc_params("waterLevelObs", base = "NGWMN")$monitoring_location_vertical_datum$description`
+#' @param water_level_above_navd88_ft
+#' `r get_ogc_params("waterLevelObs", base = "NGWMN")$water_level_above_navd88_ft$description`
+#'
+#' @param datetime
+#' `r get_ogc_params("waterLevelObs", base = "NGWMN")$sample_time$descriptiond`
+#' Multiple time_series_ids can be requested as a character vector.
```
ehinman: A little confused on this parameter. It's called `datetime` but it takes timeseries ids? Should it be called `sample_time`? Even that is confusing with the description...

ldecicco-USGS: Whoops! Nope, that's just an artifact of copy/pasting

R/read_ngwmn_lithology.R
ldecicco-USGS: Mike convinced me that it's a similar idea as the "tidy design principles":
https://design.tidyverse.org/dots-after-required.html

The idea is that the stuff before and after the dots are kind of on different levels. So in our case, the parameters after the dots are ones that I would expect are very seldom changed, and if the user does change them, they should basically be doing it at their own risk. Also, those are arguments that (basically) are specific to dataRetrieval functionality and not sent to the API. The caveat is "limit" because that is sent to the API...but most dataRetrieval users shouldn't have to worry about adjusting "limit" unless they are dealing with crappy internet or something - so that's why we put it under the dots.

R/setAccess.R
ldecicco-USGS: Nope!

ehinman (approved):
These are working well for me and are straightforward. Added some comments about how some of the input parameters seem a little bit silly without knowledge of how to use CQL2 and/or some custom code in dataRetrieval that allows users to enter ranges --  maybe in a future PR if people would find it useful.

R/read_ngwmn_well_construction.R
ldecicco-USGS: Added to the docs, and added a check in the code.

R/read_ngwmn_lithology.R
ldecicco-USGS: Someone wrote a blog about this too:
https://www.cynkra.com/blog/2026-06-12-dots/
Admittedly, we aren't using it exactly like the tidyverse folks, but kinda like them...

R/read_ngwmn_providers.R
ldecicco-USGS: added some states with links, so now we have an example that is mixed.

R/read_ngwmn_sites.R
ldecicco-USGS: I added `read_ngwmn` to take custom CQL:

```r
 cql <- '{
  "op": "between",
    "args": [
       { "property": "water_level_above_navd88_ft" },
       [ "100.00", "200.00" ]
    ]
 }'

 wl_data <- read_ngwmn(service = "waterLevelObs",
                       monitoring_location_id = c("USGS-272838082142201", 
                                                  "USGS-404159100494601", 
                                                  "USGS-401216080362703"),
                       CQL = cql)
```

maybe down the road we can think about ways to add numeric ranges like we have times...

PR #912, opened 2026-07-07
Latest field

ldecicco-USGS:
Latest field measurement endpoint is available. But, it's going to be ~2nd week of August before all the data is properly mapped. As it is the function works, but a lot of sites won't come back with data just yet. I think it's OK to push up anyway.


R/AAA.R
```diff
@@ -3,7 +3,9 @@ pkg.env <- new.env()
 .onLoad <- function(libname, pkgname) {
   suppressMessages(setAccess("public"))
   pkg.env$local_sf <- requireNamespace("sf", quietly = TRUE)
-  options("dataRetrieval.nldi_base" = "https://api.water.usgs.gov/nldi/linked-data/")
+  options(
```
jzemmels, line 6: Is this change from Air formatter? Are there other, old files in the project that haven't been similarly formatted yet and if we could do so in this PR?

R/read_waterdata_continuous.R
```diff
@@ -120,6 +120,8 @@ read_waterdata_continuous <- function(
   value = NA,
   last_modified = NA_character_,
   time = NA_character_,
+  skipGeometry = TRUE,
```
jzemmels, line 123: 😎

R/read_waterdata_latest_field.R
```diff
@@ -0,0 +1,81 @@
...
+#'                                          "time"))
+#'
+#' field_data <- read_waterdata_latest_field(monitoring_location_id = site,
+#'                            skipGeometry = TRUE)
+#'
+#' multi_site <- read_waterdata_latest_field(monitoring_location_id =  c("USGS-01435000",
+#'                                                                       "USGS-14202650"))
+#'
+#' }
+read_waterdata_latest_field <- function(
```
jzemmels: I'm wondering if there could be confusion that the associated function to this one ends in `_field_measurements`? I recognize that the super long function name is not ideal, but this goes against the naming of the other two `_latest` functions.

R/read_waterdata_ratings.R
jzemmels, line 1: Running into an error with running the examples for `read_waterdata_ratings`:

Image

R/read_waterdata_ratings.R
```diff
@@ -181,8 +190,135 @@ download_convert <- function(feature, file_path, file_type) {
...
+    other_df
+  )
+
+  final$value <- gsub("[\\x00-\\x1F\\x7F]+", " ", final$value, perl = TRUE)
+  final$value <- gsub("\\s+", " ", trimws(final$value))
+  final$header <- gsub("\\s+", " ", trimws(final$header))
+
+  final <- final[!(final$header == "" & final$value == ""), , drop = FALSE]
+  final <- unique(final)
+  rownames(final) <- final$header
```
jzemmels, line 319: This line causing an error when running examples, and is seemingly unnecessary since rownames set to NULL below.
```suggestion
```

R/read_waterdata_ratings.R
```diff
@@ -181,8 +190,135 @@ download_convert <- function(feature, file_path, file_type) {
...
+  # final_t <- data.frame(t(final))
```
jzemmels, line 320: Delete this?
```suggestion
```

jzemmels (commented):
Changes look good. One bug with the read_waterdata_ratings function and a couple comments.

R/read_waterdata_ratings.R
ldecicco-USGS: I don't know what happened, but there's some overlap between this PR and the rating curve one (I probably got a little careless with pushing up to code.usgs.gov to check on the documentation and then the various changes got contaminated). Ignore this for now and I think that's fixed on the other PR.

R/AAA.R
ldecicco-USGS: There was a PR awhile back which was a ton of Air formatting updates. Not sure why this didn't make it in back then.

PR #913, opened 2026-07-09
Extra query params

ldecicco-USGS:
* Created make_monitoring_location_arguments to help users understand 
what new optional arguments are available in many of the read_waterdata functions.
* Added ability to use `...` in the Water Data functions for: daily, latest-continuous,
field-measurements, latest-daily, latest-field-measurements, continuous, and peaks. 
The "..." argument can now accept any parameter that is output from the
make_monitoring_location_arguments function.
* Added "q" argument to read_waterdata_monitoring_location. Full-text search across
the most relevant text fields for this collection (e.g. site name, identifier,
number, state, county, and site type).
* Added "skipGeometry" argument to read_waterdata_continuous. Previously geometry
wasn't included with continuous data, but now is supported.

New examples:

```r
dane <- read_waterdata_continuous(
  state_name = "Wisconsin",
  county_name = "Dane County",
  parameter_code = "00060",
  time = "P1D")

dane <- read_waterdata_latest_daily(
  state_name = "Wisconsin",
  county_name = "Dane County",
  parameter_code = "00060",
  time = "P7D")

dane <- read_waterdata_peaks(
  state_name = "Wisconsin",
  county_name = "Dane County",
  parameter_code = "00060")

dv_dane <- read_waterdata_daily(
   state_name = "Wisconsin",
   county_name = "Dane County",
   parameter_code = "00060",
   time = "P7D")

dane <- read_waterdata_field_measurements(
  state_name = "Wisconsin",
  county_name = "Dane County",
  parameter_code = "00060",
  time = "P30D")

dane <- read_waterdata_latest_continuous(
  state_name = "Wisconsin",
  county_name = "Dane County",
  parameter_code = "00060",
  time = "P1D")

dane <- read_waterdata_latest_field_measurements(
  state_name = "Wisconsin",
  county_name = "Dane County",
  time = "P30D")

poudre <- read_waterdata_monitoring_location(q = "poudre")

```


R/read_waterdata_daily.R
```diff
@@ -97,9 +107,15 @@ read_waterdata_daily <- function(
 ) {
   service <- "daily"
   output_id <- "daily_id"
-  rlang::check_dots_empty()
 
   args <- mget(names(formals()))
+
+  args <- cleanup_arguments(
```
ehinman, line 113: Is the purpose of this function to make all of the other possible non-returned arguments NA?

ehinman: Never mind, I get it now. Forgot how `do.call` works.

R/read_waterdata_monitoring_location.R
```diff
@@ -44,10 +44,27 @@
...
+#' (e.g. site name, identifier, number, state, county, and site type).
+#' Matching rules:
+#' - Case-insensitive, with prefix matching (e.g. poud matches POUDRE).
+#' - Words separated by spaces are AND'd together (all must match).
+#' - Terms separated by commas are OR'd together (any may match).
+#' - Common abbreviations are expanded, so river also matches sites stored as RV (and vice versa).
+#' Examples are:
+#' q=poudre - sites whose name/identifier starts with or contains poudre
+#' q=colorado river - sites matching both colorado and river/rv
+#' q=poudre,cache creek - sites matching poudre OR (cache AND creek/ck)
```
ehinman, line 66: 😎

R/read_waterdata_continuous.R
```diff
@@ -31,8 +31,11 @@
 #' Available options are:
 #' `r dataRetrieval:::get_properties_for_docs("continuous", "continuous_id")`.
 #' The default (`NA`) will return all columns of the data.
-#' @param \dots Not used. Included to help differentiate official Water Data API arguments
-#' from more seldom used, optional dataRetrieval-specific arguments.
+#' @param \dots Arguments that can be queried,
+#' but are not returned. These are used as alternatives to specifying specific
+#' monitoring_location_ids. See `make_monitoring_location_arguments()`
```
ehinman: Should this say "See `?make_monitoring_location_arguments()`"?

ehinman (commented):
Neat! Love that these arguments are available now. I tested out a few and it worked really well. Some general feedback, trying to take the perspective of someone encountering this functionality for the first time:

The endpoints that take advantage of the arguments in `make_monitoring_location_arguments()` say "See make_monitoring_location_arguments() to get a list of all possible arguments available".  It's unclear whether the user should actually run the function, add in their desired inputs, or go to the help page. 

I went to the help page, which has all of the possible arguments and their definitions, but then I was thrown off by the description: "Users can use this function to create a list of possible arguments that can be used as input to the monitoring_location_arguments argument..." My first question was "how?", so I scrolled to the examples at the bottom. After running the examples, I found myself confused: would someone ever need to run those examples? The function simply creates a list of NA inputs, and replaces NA with the user's input if they populate an argument. How would the user use this to grab data? I'd suggest removing the examples at the bottom and change the description of the function to emphasize that the function is intended as a reference list with a comprehensive help page, rather than something you'd use in another function/workflow.

Given this view, I'd also suggest renaming the function to `show_monitoring_location_arguments` or `get_monitoring_location_arguments` or `reference_monitoring_location_arguments` or similar. The verb "make" makes it seem like it will create something for you to use, but I can't figure out how I'd use it besides as a reference!

ehinman:
I also played around with the `read_wateruse` function. I think the documentation on the API and models (on the actual data companion webpages, not necessarily here) could be improved for dataRetrieval/API users. I looked at the dataRetrieval examples and it took me several clicks on https://water.usgs.gov/nwaa-data/data-catalog/ to understand what the different model codes and variables were. It was kind of slow going.

Then, I ran into several 404's when trying to download data with a year-month or year that wasn't in the API. Example:

```
wu1 <- read_wateruse(model = "wu-public-supply-wd",
                      variable = c("pswdtot", "pswdgw", "pswdsw"),
                      location = "stateCd:WI",
                      startdate = "2025-01",
                      timeres = "monthly")
```
I didn't realize that this model only goes through 2020.  Overall, it would be nice to see a table of these models, their variables and variable codes, variable units, date range, time resolution, etc. That would make exploring the data a lot easier. With respect to dataRetrieval, though, I wonder if you can catch some of these requests that don't fit the model's parameters (like a start date that isn't currently supported by the model dataset) and instead return an empty dataframe rather than an error. I suppose the problem is discerning between 404's and their severity.

Lastly, and minorly, I totally read `timeres` wrong and it wasn't clear to me it's short for "time resolution". It's kind of a weird looking input! 😝

ldecicco-USGS:
I think we'll save the "read_wateruse" function for the next CRAN push and try to get these (and the NGWMN) out first.

PR #917, opened 2026-08-04
CRAN 2.7.26

ldecicco-USGS:
* Added suite of read_ngwmn to access new National Groundwater Monitoring
Network API.
* Created make_monitoring_location_arguments to help users understand 
what new optional arguments are available in many of the read_waterdata functions.
* Added ability to use `...` in the Water Data functions for: daily, latest-continuous, field-measurements, latest-daily, latest-field-measurements, continuous, and peaks. The "..." argument can now accept any parameter that is output from the make_monitoring_location_arguments function.
* Added "q" argument to read_waterdata_monitoring_location. Full-text search across the most relevant text fields for this collection (e.g. site name, identifier, number, state, county, and site type).
* Added "skipGeometry" argument to read_waterdata_continuous. Previously geometry wasn't included with continuous data, but now is supported.


R/deal_with_empty.R
```diff
@@ -25,18 +26,31 @@ deal_with_empty <- function(
...
+      )
       properties <- names(schema$properties)
     }
     return_list <- data.frame(matrix(nrow = 0, ncol = length(properties)))
     return_list <- lapply(return_list, as.character)
     names(return_list) <- properties
 
-    single_params <- c("datetime", "last_modified", "begin", "end", "time")
+    single_params <- c(
+      "datetime",
```
ehinman: `time_periods`? I see that used in `construct_api_requests.R`

ldecicco-USGS: Good call!

R/read_ngwmn_providers.R
```diff
@@ -0,0 +1,57 @@
...
+  args <- mget(names(formals()))
+
+  return_list <- get_ogc_data(
+    args = args,
+    output_id = "id",
+    service = service,
+    base = "NGWMN"
+  )
+
+  return_list <- sf::st_drop_geometry(return_list)
```
ehinman, line 54: This must be empty, huh?

ldecicco-USGS: yeah, they don't do geometry

vignettes/Status.Rmd
```diff
@@ -181,9 +181,6 @@ Currently, any "qw" data summaries provided by `whatNWISdata` are very likely ou
 
 ## National Groundwater Monitoring Network
 
-Coming soon, a new new access point for the National Groundwater Monitoring Network. The original `readNGWMNdata` functions have been broken for some time. Stay tuned!
-
-
-
+Coming soon, a new new access point for the National Groundwater Monitoring Network. The original `readNGWMNdata` functions have been broken for some time. `read_waterdata_use` is currently in development!
```
ehinman: Should this be updated to say these functions are available now?

ehinman: Also, "new" is repeated. Is that intentional?

R/setAccess.R
```diff
@@ -76,6 +76,5 @@ Please contact comptools@usgs.gov for more information."
   pkg.env$samplesData <- "https://api.waterdata.usgs.gov/samples-data/summary"
   pkg.env$status <- "https://www.waterqualitydata.us/wqx3/status/"
 
-  pkg.env$NGWMN <- "https://cida.usgs.gov/ngwmn_cache/sos"
```
ehinman, line 79: No replacement needed?

R/readNWISunit.R
```diff
@@ -454,18 +454,17 @@ readNWISstat <- function(
...
+readNWISuse <- function(stateCd,
+                        countyCd,
+                        years = "ALL",
+                        categories = "ALL",
+                        convertType = TRUE,
+                        transform = FALSE) {
+
+  .Deprecated(new = "read_waterdata_use",
+              package = "dataRetrieval", 
+              msg = "NWIS servers for water use have been decommission. New functions are being developed.")
```
ehinman: Should this be undone since `read_waterdata_use` doesn't exist in this version?

R/setAccess.R
ldecicco-USGS: Correct, we're phasing this out

R/get_monitoring_location_arguments.R
```diff
@@ -0,0 +1,204 @@
+#' Get Monitoring Location Arguments
+#'
+#' Many read_waterdata functions have a long list of arguments that can be used
+#' to find sites that have data. Users can use this function to create a list
+#' of possible arguments that can be used as input to the `monitoring_location_arguments`
+#' argument (found in many of the read_waterdata functions). This function
+#' also is used to check that user supplied available parameters to their queries
+#' when using the `monitoring_location_arguments` argument.
```
ehinman:
```suggestion
#' when using the `get_monitoring_location_arguments` argument.
```

R/get_monitoring_location_arguments.R
```diff
@@ -0,0 +1,204 @@
+#' Get Monitoring Location Arguments
+#'
+#' Many read_waterdata functions have a long list of arguments that can be used
+#' to find sites that have data. Users can use this function to create a list
+#' of possible arguments that can be used as input to the `monitoring_location_arguments`
```
ehinman:
```suggestion
#' of possible arguments that can be used as input to the `get_monitoring_location_arguments`
```

R/get_monitoring_location_arguments.R
ehinman: I'm still a little bit confused about how someone would use this function beyond as a reference. Are you suggesting that they feed the list created by the function to something like `get_daily`? What if they also wanted to specify a stat code or something? They'd just add that input to the list?

ldecicco-USGS: I feel like the function has 2 reasons to exist... but being a reference is the main one:

1. It gets all the available parameter's descriptions (so, reference)
2. It's used to check if the parameters are actually allowed in the specific endpoint. 

So if a user said `statename = "Wisconsin"`, they'd see:

```r
 dv_data_sf <- read_waterdata_daily(statename = "Wisconsin",
                                    parameter_code = "00060",
                                    time = c("2021-01-01", "2022-01-01"))
Error:
! `...` must be empty.
✖ Problematic argument:
• statename = "Wisconsin"
Run `rlang::last_trace()` to see where the error occurred.
```
We get that error message because `read_waterdata_daily` called `get_monitoring_location_arguments` to check that the arguments were legit.

Using the correct state_name works:
```r
> dv_data_sf <- read_waterdata_daily(state_name = "Wisconsin",
+                                    parameter_code = "00060",
+                                    time = c("2021-01-01", "2022-01-01"))
Requesting:
https://api.waterdata.usgs.gov/ogcapi/v0/collections/daily/items?f=json&lang=en-US&parameter_code=00060&time=2021-01-01%2F2022-01-01&state_name=Wisconsin&limit=50000
```

So if a user was confused why "statename" doesn't work, we'd send them to the help page for `get_monitoring_location_arguments` to see that it should be state_name.

Does that help?

ehinman: Should the user know that these functions are calling `get_monitoring_location_arguments()` to check inputs under the hood? The error message doesn't say anything about `get_monitoring_location_arguments()`, but I suppose they should have found `state_name` from the `get_monitoring_location_arguments()` page? Maybe what would clear this up for me is to edit the description to say "This function is also used by other functions to check that the user supplied available parameters to their queries".

ldecicco-USGS: The description for `get_monitoring_location_arguments()`? Or the description in the `...` argument in the individual read_waterdata functions?

ehinman: The description for `get_monitoring_location_arguments`.

ehinman (approved):

R/readNWISunit.R
```diff
@@ -454,17 +454,17 @@ readNWISstat <- function(
...
+  stateCd,
+  countyCd,
+  years = "ALL",
+  categories = "ALL",
+  convertType = TRUE,
+  transform = FALSE
+) {
+  .Deprecated(
+    package = "dataRetrieval",
+    msg = "NWIS servers for water use have been decommission. New functions are being developed."
```
ehinman: "decommissioned"

